PR #13, opened 2020-03-03
Add header pic, change favicon and title

huayunh:
- Add title picture in the drawer header
- Change title and favicon
- Adjust background image positioning
- global yarn prettier & lint


emclaug2:
If you refresh the doc site on an API story, the icon is no longer set.  In api.stories.ts, if you move the updateTitle() call to always run, this will fix the problem.

huayunh:
> If you refresh the doc site on an API story, the icon is no longer set. In api.stories.ts, if you move the updateTitle() call to always run, this will fix the problem.

Thanks. My bad.

jeffvg:
Storybook favicon.ico served up and displays for a moment then our pxblue icon displays

PR #26, opened 2020-03-27
Convert Storybook to use Component Story Format

### LINKED ISSUE #14: Update Storybook Stories to use CSF
Update our storybook stories to Component Story Format. Storybook 5.2+ recommends CSF for increased code modularity & other perks.

There is a branched called 'experiment/csf-future-enhancement' that attempts the transition from withStories to CSF, but I could not complete this feature.  Documentation for CSF in Angular is very weak and there is little support.  I have linked an issue in Storybook that discusses the lack of documentation.

We are use the withStories(...) API.

No

emclaug2:
Fixes #14 

- Use CSF format, split stories into modular files. 
- Storysource is cleaner now for stories.
- Individual stories work the same as before.


jeffvg:
Is this a work around/fix for CSF issue until SB release the fix in 6.0? What will happen if angular SB is upgraded later?

joebochill:
What do you mean when you say it 'hides module info'?

emclaug2:
> What do you mean when you say it 'hides module info'?

Here is a basic file that would be seen in the story source plugin:

`
import { text } from '@storybook/addon-knobs';

export const withBasicConfig = (): any => ({
    template: `
          <pxb-hero [label]="label">
             <i primary class="pxb-grade_a"></i>
          </pxb-hero>
      `,
    props: {
        label: text('label', 'Efficiency'),
    },
});
`

The original storysource would include this snippet in additional to the module declaration & all other stories source code.

emclaug2:
> Is this a work around/fix for CSF issue until SB release the fix in 6.0? What will happen if angular SB is upgraded later?

This PR is converting our stories into the recommended Component Story Format.  When storybook is updated to 6, I don't think CSF will be deprecated before `withStories` API, since CSF was introduced in 5.2+. 

The CSF issue that I mentioned before is talking about how the docs for converting an Angular SB project to use CSF is really bad.  CSF is supported for Angular projects, but it's just unclear what the best practices are at this point; or how to do the migration.  This PR is the result of a lot of trial and error.

PR #29, opened 2020-03-31
Feature/scorecard

emclaug2:
- Add scorecard component
- Add examples to Showcase & Storybook
- Add configurable background color to HeroComponent
- Add a FontSize type to the HeroComponent
- Created an issue to make SVGs easier to integrate into the HeroComponent

Edits:
- Added a testUtil to search dom tree, count instances of selector.
- Enforce Scorecard class names as if they are API props.


jeffvg:
missing  actionLimit and headerBackgroundImage prop?

components/src/core/hero/hero.component.ts
```diff
@@ -23,24 +26,24 @@ import { Component, Input, OnInit } from '@angular/core';
...
-        // We can't support dynamic iconSize w/ px-blue icons until https://github.com/angular/components/issues/5188 is fixed
+    // We can't support dynamic iconSize w/ px-blue icons until https://github.com/angular/components/issues/5188 is fixed
+    ngOnChanges(): void {
         this.iSize =
             this.iconSize === 'large'
                 ? 72
                 : this.iconSize === 'normal'
                 ? 36
-                : (this.iSize = Math.max(10, Math.min(72, parseInt(this.iconSize, 10))));
+                : this.iconSize;
```
joebochill: should we allow this at all if it doesn't work with PX Blue icons? Or just stick to the two default options?

components/src/core/score-card/public-api.ts
```diff
@@ -0,0 +1,2 @@
+export * from './scorecard.component';
+export * from './scorecard.module';
```
joebochill: Question: why does Hero only export the module, while all of the other components export the module and the component?

components/src/core/score-card/scorecard.component.scss
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// Styles applied to supplied action items.
+:host ::ng-deep .action-items * {
+    margin-left: 12px;
+    cursor: pointer;
+}
+
+.header {
+    border-top-left-radius: 4px;
```
joebochill: is this a roundness that exists in the theme?

components/src/core/score-card/scorecard.component.ts
```diff
@@ -0,0 +1,43 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+    selector: 'pxb-scorecard',
+    template: `
+        <mat-card class="root">
+            <div class="header" [style.backgroundColor]="headerColor" [style.color]="headerFontColor">
+                <div class="header-overlay"></div>
```
joebochill: can/should we expose backgroundImage as a prop for convenience so a user doesn't have to make a bunch of CSS rules for all of their cards?

components/src/core/score-card/scorecard.component.ts
```diff
@@ -0,0 +1,43 @@
...
+                    </div>
+                    <div class="action-items">
+                        <ng-content select="[action-items]"></ng-content>
+                    </div>
+                </div>
+            </div>
+            <mat-card-content>
+                <div class="body">
+                    <ng-content select="[body]"></ng-content>
+                    <div class="badge-wrapper" [style.marginTop.px]="badgeOffset || 'inherit'">
```
joebochill: these wrappers still render even if no badge, etc. is supplied...how do we feel about that?

demos/showcase/src/app/app.component.html
```diff
@@ -1,7 +1,96 @@
+<div style="display: flex; padding: 16px; justify-content: space-between">
+    <pxb-scorecard
+        [headerTitle]="'Substation 3'"
+        [headerSubtitle]="'High Humidity Alarm'"
+        [headerInfo]="'4 Devices'"
+        [headerColor]="colors.red[500]"
+        [headerFontColor]="colors.white[50]"
+    >
+        <mat-icon action-items>more_vert</mat-icon>
+        <mat-list body>
```
joebochill: Can you kill the extra top padding on all of these lists...particularly the actionRow ones since they mess up the vertical alignment

demos/showcase/src/app/app.component.html
```diff
@@ -1,7 +1,96 @@
+<div style="display: flex; padding: 16px; justify-content: space-between">
+    <pxb-scorecard
```
joebochill: can you drop a background image on these to match the react showcase

demos/showcase/src/app/app.component.html
```diff
@@ -1,7 +1,96 @@
+<div style="display: flex; padding: 16px; justify-content: space-between">
```
joebochill: can we do flex-start instead of space-between to match react

components/src/core/hero/hero.component.ts
```diff
@@ -1,19 +1,22 @@
...
 @Component({
     selector: 'pxb-hero',
     template: `
-        <div class="wrapper">
-            <span
-                class="icon"
+        <div class="root">
+            <div
+                class="primary-icon"
+                [style.backgroundColor]="iconBackgroundColor"
```
joebochill: this backgroundColor doesn't have any padding around icons that are close to the edge of the 24x24 grid:
image

demos/storybook/stories/ hero/with-full-config.stories.ts
```diff
@@ -13,6 +13,8 @@ export const withFullConfig = (): any => ({
         value: text('value', '470'),
         units: text('units', 'RPM'),
         showSecondary: boolean('Show Secondary Icon', true),
-        iconColor: color('primary.style.color', Colors.blue[500]),
+        iconSize: number('iconSize', 98),
```
joebochill: should we just limit this to large and normal?

docs/ScoreCard.md
```diff
@@ -0,0 +1,55 @@
...
+...
+imports: [
+    ScoreCardModule
+  ],
+...
+```
+
+```typescript
+// Example with basic body content and header action items
+<pxb-scorecard [headerTitle]="'Title'" >
```
joebochill: Can we show an example that uses badge like react since that's the most complicated and not terribly intuitive

docs/ScoreCard.md
```diff
@@ -0,0 +1,55 @@
...
+
+<div style="align-items: center; text-align: center; width: 100%; display: flex; justify-content: space-evenly; margin-bottom: 20px">
+    <img width="40%" alt="Score Card with multiple highlighted values" src="./images/scoreCard.png"><br/>
+    <img width="40%" alt="Score Card with single badge-style value" src="./images/scoreCard_alt.png"><br/>
+</div>
+
+## Usage
+
+<div style="align-items: center; text-align: center; width: 100%; display: flex; justify-content: space-evenly; margin-bottom: 20px">
+    <img width="80%" alt="Score Card with labels on different parts" src="./images/scoreCardAnatomy.png"><br/>
```
joebochill: you'll need a new image for anatomy since the API/usage differs from react

huayunh: let me know when the API is properly tuned and I can push new pics to this branch

joebochill: thanks @huayunh

emclaug2:
> missing actionLimit and headerBackgroundImage prop?

The `action-items` is not an array like in the React ScoreCard, it's a projected `ng-content` which could be an array or a single element, so I left `actionLimit` out.

I'm glad you brought up the missing `backgroundImage` prop.  We're going to have a meeting today where we discuss style props in general.  I think we can actually remove all style props and if the user wants to apply custom styling, they can do so via sass.  

This way we never have to pick & choose which props we want to expose, it makes our API simpler while allowing everything to be customizable via sass.  This is how material UI does it as well.

components/package.json
```diff
@@ -14,10 +14,12 @@
     "lint:fix": "eslint \"src/**/**.ts\" --fix"
   },
   "peerDependencies": {
-    "@pxblue/colors": "^2.0.0",
-    "@angular/material": "^8.2.3"
+    "@pxblue/angular-themes": "^4.0.0",
```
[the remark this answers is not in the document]
emclaug2: I've removed this dependency; we should be able to theme this component after the theme spike is completed.

components/src/core/hero/hero.component.spec.ts
```diff
@@ -13,11 +13,11 @@ describe('HeroComponent', () => {
         }).compileComponents();
     }));
 
-    it(`Div element should have class as 'wrapper'`, () => {
+    it(`root element should have class as 'root'`, () => {
         const fixture = TestBed.createComponent(HeroComponent);
         const heroComponent = fixture.debugElement;
         const wrapperDiv: HTMLElement = heroComponent.query(By.css('div')).nativeElement;
-        expect(wrapperDiv.getAttribute('class')).toEqual('wrapper');
+        expect(wrapperDiv.getAttribute('class')).toEqual('root');
```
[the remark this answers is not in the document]
emclaug2: Unless we are protecting our class names from changes at this stage, I would say they provide no value. 

I think it would be a good idea when we release R16 that we treat our class names the same way as we do our @Input names.  We should not be changing them very often.

components/src/core/hero/hero.component.ts
emclaug2: I wanted to match the flexibility of the React HeroComponent; it allows an icon of any size.  I'm personally not a fan of just allowing for two sizes; it's good to have them as options but not as a limitation.  I made a defect to address this limitation.

components/src/core/score-card/public-api.ts
emclaug2: This is actually a bug in the Hero public-api.  I've updated it to export all from `hero.component` and `hero-banner.component` as well.  The reasoning for this is to allow the `HeroComponent` and `HeroBannerComponent` to be propagated upwards so a user can simply `import { HeroComponent } from '@pxblue/angular-components'`.

demos/showcase/src/app/app.component.html
emclaug2: Sure thing, I've updated it to match.

components/src/core/hero/hero.component.spec.ts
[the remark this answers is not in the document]
joebochill: yes, if we are expecting users to do styles based on these classes, they should almost never be changed...any changes would be considered breaking

components/src/core/hero/hero.component.ts
joebochill: There is an issue for this already on the icons repo (https://github.com/pxblue/icons/issues/27). I'm not sure we should support this until it actually works...if a user tries to change the size they're going to see this:
```
<pxb-hero [label]="'Temperature'" [value]="'98'" [units]="'°F'" [iconSize]="100">
    <mat-icon svgIcon="px-icons:temp" primary></mat-icon>
</pxb-hero>
<pxb-hero [label]="'Humidity'" [value]="'54'" [units]="'%'" [iconSize]="100">
    <mat-icon mat-list-icon primary>cloud</mat-icon>
</pxb-hero>
```
image

components/src/core/hero/hero.component.spec.ts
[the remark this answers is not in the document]
joebochill: are you also planning to prefix these classes (e.g., `pxblue-root`) like we discussed previously?

components/src/core/hero/hero.component.scss
```diff
@@ -11,49 +11,57 @@
...
     font-size: 100%;
-    line-height: 1.2;
-    text-align: center;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    white-space: nowrap;
+}
+
+:host ::ng-deep .primary-icon i {
+    margin-top: -1.5px;
```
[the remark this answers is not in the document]
emclaug2: This is a hack that I used to center align some icons; this shouldn't have been left it.  I'll remove it, but I'm still having some trouble getting some of the icons to be center aligned whenever there's a background applied.

components/src/core/hero/hero.component.spec.ts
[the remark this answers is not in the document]
emclaug2: I think it's a good idea, I'll prefix all all the class names.

demos/storybook/stories/ hero/with-full-config.stories.ts
emclaug2: I'm going to update to the iconSize to only support two values 'normal' and 'large'.  If a user wants a custom size they can override it via sass.

components/src/core/hero/hero.component.ts
emclaug2: I have to check to see if this isn't an issue in the React library too.

emclaug2: image

This is from the react library.  If a user is trying to use an icon that requires additional padding, I think they can just add the padding themselves.  Most icons dont stretch all the way vertically or horizontally, so I think Cloud is in the minority here.

components/src/core/hero/hero.component.scss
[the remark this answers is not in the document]
emclaug2: I fixed the vertical alignment.

docs/ScoreCard.md
emclaug2: Yep updated with a more complicated example; also they should be using the storybook story-source as an alternative source for complicated examples.

demos/storybook/.editorconfig
```diff
@@ -0,0 +1,13 @@
+# Editor configuration, see http://editorconfig.org
```
huayunh, line 1: the editor config sneaked in

demos/storybook/package.json
```diff
@@ -8,10 +8,8 @@
     "storybook": "start-storybook -p 6006 -s ../../docs/,./assets,./node_modules/typeface-open-sans/files",
     "build-storybook": "build-storybook -s ../../docs/,./assets,./node_modules/typeface-open-sans/files",
     "lint": "eslint \"stories/**/**.ts\"",
-    "lint:fix": "eslint \"stories/**/**.ts\" --fix",
-    "prettier": "prettier \"stories/**/**.{ts,json,css,scss,html}\" --write"
+    "lint:fix": "eslint \"stories/**/**.ts\" --fix"
   },
-  "prettier": "@pxblue/prettier-config",
```
huayunh, line 14: Why was prettier removed from the storybook?

emclaug2: I removed the prettier config for the angular storybook because most of the code in these stories is not JS or TS, it's actually just a big `template` string.   Whenever I ran prettier on this large string, I would lose a lot of the formatting that I did manually, so I swapped it with an editorconfig file that'll keep our tab indentations consistent at 4 spaces.

demos/storybook/.editorconfig
emclaug2: This was an intentional add; this is actually the .editorconfig we have in our @pxblue/prettier-config package.  It finally has a use-case.

components/src/core/hero/hero.component.ts
joebochill: It's not just cloud...it's pretty much any square-shaped icon too. But since it's also an issue in React, we don't have to deal with it right now.

components/src/core/score-card/scorecard.component.ts
```diff
@@ -0,0 +1,41 @@
...
+                <div class="pxb-body">
+                    <ng-content select="[body]"></ng-content>
+                    <div class="pxb-badge-wrapper" [style.marginTop.px]="badgeOffset || 'inherit'">
+                        <ng-content select="[badge]"></ng-content>
+                    </div>
+                </div>
+                <mat-divider></mat-divider>
+                <div class="pxb-action-row">
+                    <ng-content select="[action-row]"></ng-content>
+                </div>
```
joebochill, line 30: there should not be a divider if no action-row is provided

joebochill: image

demos/storybook/package.json
joebochill: would this be resolved by splitting the template string into a separate html file?

demos/storybook/stories/scorecard/with-score-badge.stories.ts
```diff
@@ -0,0 +1,63 @@
...
+import { action } from '@storybook/addon-actions';
+import { withCustomHeaderStyles } from './with-custom-header.stories';
+import { demoActions } from './with-actions.stories';
+import { ViewEncapsulation } from '@angular/core';
+
+export const withScoreBadge = (): any => ({
+    styles: [
+        `${withCustomHeaderStyles}
+        .sb-scorecard-content mat-list-item {
+            height: 36px!important;
```
joebochill, line 12: are we anticipating users having to use a lot of `!important` rules to customize our components?

components/src/core/hero/hero.component.ts
emclaug2: I've reverted the hero back to accepting two icon sizes.  I want to revisit this in the future so that either more sizes are supported or custom sizing is possible.

components/src/core/score-card/scorecard.component.ts
emclaug2: I don't think we should expose backgroundImage as a prop.  What if the user wants to change the opacity, size, etc.  I'll argue that most style props should be handled by our classes.

components/src/core/score-card/scorecard.component.scss
emclaug2: I've updated this to inherit the border roundness from the mat card container.

components/src/core/score-card/scorecard.component.ts
emclaug2: Since it's an empty div it shouldn't cause any issues.  Do you have any concerns about this?

emclaug2: https://github.com/angular/angular/issues/26083#issuecomment-425227938 . This is an alternative.

joebochill: My only concern is if the empty containers have styling associated with them (e.g., margins) that would affect the positioning of other elements.

components/src/core/score-card/scorecard.component.ts
emclaug2: Good catch, I get to try out the conditional based on the presence of a `ng-content` being provided now.

demos/storybook/stories/scorecard/with-score-badge.stories.ts
emclaug2: This example you've pointed out is styling the material list item that is passed into our component.   Our components will have the same limitation if the user does not use enough detail in their css selectors. 

`.red-scorecard {
    .pxb-root .pxb-header {
        background-color: map-get($pxb-red, 500);
    }
}` 

This is an example of the showcase app overwriting the default header background color.  No !important needed.

emclaug2:
> * Can we alphabetize the components in the storybook
> * Remove top padding from lists in the showcase (body and footer)
> * Hero README should have an anatomy image (Huayun?)
> * Are we going to prefix the classes in the existing components too?
> * Remove the Available properties section of Hero Readme
> * Should we update the API tables to say 'attributes' instead of 'props'?
> * Do we need a default column in the API tables?
> * HeroBanner usage example should have heroes in it
> * Icons in the hero stories looks smaller than should be for default
> * ChannelValue still says it still takes a color attribute for the font (Readme)
> * ChannelValue default font color looks incorrect (too black) in storybook
> * ChannelValue 'with extra large font size' -> 'with fontSize'
> * EmptyState readme needs anatomy image
> * That margin below the title for empty still bothers me

This is massive scope creep.  Updating EmptyState/ChannelValue has nothing to do with the Scorecard and is better left as separate stories.  Scorecard should be setting the example for class name, unit tests enforcing class names, API, readME format, etc.  Updating everything to match is a separate effort.

components/src/core/score-card/scorecard.component.ts
emclaug2: Updated this.

joebochill:
> > * Can we alphabetize the components in the storybook
> > * Remove top padding from lists in the showcase (body and footer)
> > * Hero README should have an anatomy image (Huayun?)
> > * Are we going to prefix the classes in the existing components too?
> > * Remove the Available properties section of Hero Readme
> > * Should we update the API tables to say 'attributes' instead of 'props'?
> > * Do we need a default column in the API tables?
> > * HeroBanner usage example should have heroes in it
> > * Icons in the hero stories looks smaller than should be for default
> > * ChannelValue still says it still takes a color attribute for the font (Readme)
> > * ChannelValue default font color looks incorrect (too black) in storybook
> > * ChannelValue 'with extra large font size' -> 'with fontSize'
> > * EmptyState readme needs anatomy image
> > * That margin below the title for empty still bothers me
> 
> Updating EmptyState/ChannelValue has nothing to do with the Scorecard and is better left as separate stories. Scorecard should be setting the example for class name, unit tests enforcing class names, API, readME format, etc. Updating everything to match is a separate effort.

Fair enough...I'll add the out of scope items to PXBLUE-1221

emclaug2:
https://github.com/pxblue/angular-component-library/pull/29#discussion_r402931856

Yes but then the storysource plugin in storybook would be useless because it does not show the most important code.  The template is the MVP in our stories shown.

emclaug2:
> * Can we alphabetize the components in the storybook
> * Remove top padding from lists in the showcase (body and footer)
> * Should we update the API tables to say 'attributes' instead of 'props'?
> 
> UPDATE: moved out of scope items to a separate story

Thanks for moving most of the stuff into a separate story.  
I think `attributes` is a good name.  I've updated it.

I'll work on the other two things now.

demos/storybook/package.json
emclaug2: Yes but then the storysource plugin in storybook would be useless because it does not show the most important code. The template is the MVP in our stories shown.

emclaug2: If there is a way to preserve the template as a non-string in the same file as what is rendered in the storysource plugin, we can keep prettier.  I don't think it is unreasonable for us to be responsible for manual code indentation since we have the editorconfig to help us.

components/src/core/score-card/scorecard.component.ts
emclaug2: This particular instance doesn't have any styles on it so I think it should be fine.  If we have to make this a conditional render, we can go with the same approach used to conditionally render the mat-divider if there's no actionRow.

Issue #30, opened 2020-04-02
Component Library Not working with AOT-enabled projects


emclaug2:
#### Describe the bug
Our component library is not working with projects that have enabled ahead of time compilation.

#### What is the expected behavior?
Library should be able to integrate into projects with or without aot enabled.

#### What are the steps to reproduce?
Run the showcase project using `ng serve --aot=true`. 

#### What is your environment?

 Framework
- [x] Angular
- [ ] React
- [ ] Ionic
- [ ] React Native

 Browser
- [ ] Internet Explorer
- [ ] Edge
- [ ] Safari
- [ ] Chrome
- [ ] Firefox

 Operating System
- [ ] MacOS
- [x] Windows
- [ ] Other

PR #32, opened 2020-04-03
Feature/aot compatibility

### LINKED ISSUE #31: Angular 9 Support
Our components do not work in a project that uses angular v9.

This is an excerpt from the angular changelogs: 
Deprecations
Importing directly from the root @angular/material entry-point is deprecated. You should instead import from specific entry-points, e.g. @angular/material/button. This aligns with @angular/cdk, makes clear where symbols originate, and helps safeguard against including unused code. The deprecated style will be removed in Angular v9.

Mat-Divider is not recognized in the Hero Component because the root import has been deprecated.  Angular 9 will not load our component library.

No

emclaug2:
Fixes #31, #30 

Adds NG 9 compatibility.
Adds NG 7/8 AOT compatibility.

Loses ability to import from nested module, ie `import { HeroModule } from '@pxblue/angular-components/core/hero'`.
This is something that we originally retained for backwards compatibility with the older angular library; this is something that angular material has started doing as well; they've actually deprecated root-level imports.   At this moment I'm not sure how to support both root level and nested imports for our library now that the index.ts files were removed.  

Angular really wants us to use the public_api.ts as an entryPoint into our lib.  Updating our entrypoint solved the AOT issues in the Showcase app.   

Converting all angular/material imports to use a nested import instead of the root-level import fixed angular 9 compatibility issues.


joebochill:
So which way are we going? Changing to the public API approach, or deprecating root level imports? I'm more inclined to follow in the footsteps of Angular Material.

Also, do these changes retain backward compatibility with Angular 7 & 8?

emclaug2:
> So which way are we going? Changing to the public API approach, or deprecating root level imports? I'm more inclined to follow in the footsteps of Angular Material.
> 
> Also, do these changes retain backward compatibility with Angular 7 & 8?

I think we should also follow in the steps of Angular Material and deprecate root level imports.  Unfortunately I'm going to have to do a little digging and modify this PR.

Our component library is made with angular 8, integrated into Showcase @ v7 & design pattern mono @ v9.  We should be fine.

PR #34, opened 2020-04-07
Feature/theme toggle

daileytj:
Fixes PXBLUE-1149.

- Add dark theme to storybook
- add theme toggle to storybook


demos/storybook/src/utils.ts
```diff
@@ -76,6 +76,11 @@ export const getReadMeStory = (): any => {
 };
 getReadMeStory.story = { name: README_STORY_NAME };
 
+export const isDarkMode = () => {
```
emclaug2, line 79: Can be simplified to:
` return darkModeLocalStorage.current === 'dark'; `

demos/storybook/.storybook/preview-head.html
```diff
@@ -10,6 +10,5 @@
     flex-direction: column;
     align-items: center;
     justify-content: center;
-    background: #efefef;
```
emclaug2, line 13: Can we preserve the slight offwhite background for the canvas?  I really think it helps the components standout for light theme.  

image


vs 

image

demos/storybook/src/welcome/welcome.component.ts
```diff
@@ -8,12 +8,15 @@ import {hideTopBanner, showTopBanner, UtilModule} from "../utils";
...
 const icon = require('../../assets/pxb-icon.svg');
+import {CommonModule} from '@angular/common';
+import { isDarkMode } from '../utils';
 
 @Component({
     selector: 'welcome',
     styleUrls: ['welcome.component.css'],
     template: `
-        <div class="pxb-blue mat-typography root"
+        <div class="mat-typography root"
```
emclaug2, line 18: Is dark mode required for the welcome story?

demos/storybook/.storybook/preview-head.html
andrupu: I agree with Evan.  The off-white (darker) background makes the component standout in light theme.

huayunh:
> > image
> > image
> 
> Should the component examples like the empty state Evan commented on be converted to dark theme as well (to get the light brackground off and better mesh with the dark themed page) or is that out of scope?

It's a bit hard to do with just markdown, I think.

demos/storybook/.storybook/preview-head.html
daileytj: Setting the background-color here overrides the theme so when we go to dark mode that canvas color persists. Could possibly use javascript to apply the styles based on the theme if we really want this.

demos/storybook/src/welcome/welcome.component.ts
daileytj: probably not... I'll remove it

demos/storybook/.storybook/preview-head.html
emclaug2: Maybe do an inline style for light theme.

jeffvg:
difference between react and angular on dark mode..gray 300 & 500

demos/storybook/.storybook/preview-head.html
huayunh: Or we can use theme.palette.background.default for dark and light theme respectively. Gray50 for light and BlackA200 for dark. 

Gray50 is a typical color used when there are some UI elements that need to "pop-out" and demands a higher elevation. Specially card-like elements.

joebochill:
> difference between react and angular on dark mode..gray 300 & 500
> 

That's an issue in the component itself (none of the angular components are properly themed in dark mode) rather than the dark theme for storybook.

demos/storybook/.storybook/config.js
```diff
@@ -30,6 +30,14 @@ pxblueTheme.brandTitle = 'PX Blue Angular Component Library';
 pxblueTheme.brandImage = require('../assets/pxblue angular.svg');
```
joebochill, line 30: can you rename this image file so it doesn't have a space in it?

demos/storybook/.storybook/config.js
```diff
@@ -30,6 +30,14 @@ pxblueTheme.brandTitle = 'PX Blue Angular Component Library';
...
 pxblueTheme.brandUrl = 'https://pxblue.github.io';
 
+pxblueDarkTheme.brandTitle = 'PX Blue Angular Component Library';
+pxblueDarkTheme.brandImage = require('../assets/pxblue angular.svg');
+pxblueDarkTheme.brandUrl = 'https://pxblue.github.io';
```
joebochill, line 35: does this work:
```
pxblueTheme.<property> = pxblueDarkTheme.<property> = <value>
```

huayunh:
### Theme

* Storybook's search bar (under the menu header pic) is changing its width upon toggling
* Enhancement issue: under the dark mode, clicking on the "Open canvas in a new tab" button on the top right will change the canvas theme back to the light mode, but not the storybook theme. 

### WRT some readme files

* Channel Value and Hero's API table did not escape the HTML characters properly. 
* Hero has both "Available properties" and "Hero API" table. Remove the first one.
* Hero API table is missing the "default" column 
* Hero Banner API table has an "example" column. We don't need it.

emclaug2:
> ### Theme
> * Storybook's search bar (under the menu header pic) is changing its width upon toggling
> * Enhancement issue: under the dark mode, clicking on the "Open canvas in a new tab" button on the top right will change the canvas theme back to the light mode, but not the storybook theme.
> 
> ### WRT some readme files
> * Channel Value and Hero's API table did not escape the HTML characters properly.
> * Hero has both "Available properties" and "Hero API" table. Remove the first one.
> * Hero API table is missing the "default" column
> * Hero Banner API table has an "example" column. We don't need it.

These readme edits are good suggestions but should be done in a separate PR.

joebochill:
> > ### Theme
> > 
> > * Storybook's search bar (under the menu header pic) is changing its width upon toggling
> > * Enhancement issue: under the dark mode, clicking on the "Open canvas in a new tab" button on the top right will change the canvas theme back to the light mode, but not the storybook theme.
> > 
> > ### WRT some readme files
> > 
> > * Channel Value and Hero's API table did not escape the HTML characters properly.
> > * Hero has both "Available properties" and "Hero API" table. Remove the first one.
> > * Hero API table is missing the "default" column
> > * Hero Banner API table has an "example" column. We don't need it.
> 
> These readme edits are good suggestions but should be done in a separate PR.

I had the same comments on Evan's PR wrt readmes. There is a separate story for cleaning those up

joebochill:
Switching between light and dark mode when you go from Readme to story is very jarring (at least for me). Also, the light gray background on the canvas is not going to allow the dark components to properly pop. If we can find a way to use the proper default background color based on the theme, that would be excellent ("Gray50 for light and BlackA200 for dark").

PR #35, opened 2020-04-07
Feature/any size hero

### LINKED ISSUE #28: Hero Component Difficult to Work with PX Blue SVGs
It's difficult to work with the Hero Component when I want to use a PX Blue SVG.
I want to accept a fontSize number that scales my icon appropriately. 

I am forced to use either 'normal' or 'large' size.

No

#### Additional Context
Non-SVG icons don't have this issue and can accept a number. 
There is an example of this in the storybook app and showcase app.

emclaug2:
Fixes #28 

- Allows number as `iconSize`
- Prefix Hero class names with `pxb-`
- Add new story demonstrating icon image formats.
- Add padding around hero icon


components/src/core/hero/hero.component.ts
```diff
@@ -1,41 +1,119 @@
...
+                [style.width.px]="iSize"
+                [style.height.px]="iSize"
+                [class.pxb-svgIcon]="hasMatSvgIcon"
             >
                 <ng-content select="[primary]"></ng-content>
             </div>
-            <span class="channel-value" [style.fontSize.rem]="fontSize === 'small' ? '1' : '1.25'">
-                <ng-content *ngIf="value === undefined" select="pxb-channel-value"></ng-content>
+            <span class="pxb-channel-value-wrapper" [style.fontSize.rem]="fontSize === 'small' ? '1' : '1.25'">
+                <ng-content select="pxb-channel-value"></ng-content>
```
joebochill, line 31: does removing this ngif check mean you can pass config through the value attribute and as children? We only want to allow one or the other at a time.

emclaug2: Yeah this makes sense, I added the conditional back.

PR #36, opened 2020-04-09
Feature/info list item forked

emclaug2:
- Add InfoListItem component
- Add Spacer component
- Update Showcase demo to use InfoListItem
- Update storybook webpack config to match React's.


emclaug2:
There's a bug in the Divider, partial not working as expected.  Will address tomorrow.

components/src/core/info-list-item/info-list-item.component.ts
```diff
@@ -0,0 +1,58 @@
...
+export class InfoListItemComponent {
+    @Input() title: string;
+    @Input() subtitle: string | any[];
+    @Input() subtitleSeparator = '\u00B7';
+    @Input() statusColor: string;
+    @Input() chevron = false;
+    @Input() dense = false;
+    @Input() hidePadding = false;
+    @Input() wrapSubtitle = false;
+    @Input() wrapTitle = false;
```
joebochill, line 49: The wrapping behavior is backwards in the storybook (wrap true makes it ellipse, wrap false makes it wrap)

docs/InfoListItem.md
```diff
@@ -0,0 +1,42 @@
...
+<pxb-info-list-item title="Status" divider="full" [statusColor]="colors.green[500]">
+    <mat-icon [style.color]="colors.green[500]" icon>eco</mat-icon>
+</pxb-info-list-item>
+```
+
+## API
+
+<div style="overflow: auto;">
+
+| Prop Name         | Description                                      | Type                                               | Required | Default        |
```
joebochill, line 26: I believe we're calling this column attributes for angular components now

docs/Spacer.md
```diff
@@ -0,0 +1,37 @@
...
+<pxb-spacer flex="1"></pxb-spacer>
+// using pixels
+<pxb-spacer width="50"></pxb-spacer>
+```
+
+## API
+
+<div style="overflow: auto;">
+
+| Prop Name | Description                             | Type     | Required | Default |
```
joebochill, line 31: 'attributes'

docs/Spacer.md
```diff
@@ -0,0 +1,37 @@
...
+|-----------|-----------------------------------------|----------|----------|---------|
+| flex      | Flex grow/shrink value for flex layouts | `number` | no       | 0       |
```
joebochill, line 33: default should be 1

joebochill:
Our storybook bundle is also getting quite large...we may want to start thinking about how we can optimize this, especially if we're planning to add a bunch of new components:
image

components/src/core/info-list-item/info-list-item.component.ts
```diff
@@ -0,0 +1,58 @@
...
+            [class.pxb-status]="statusColor"
+            [style.borderLeftColor]="statusColor"
+        >
+            <div mat-list-icon class="pxb-icon" [class.pxb-hide-padding]="hidePadding">
+                <ng-content select="[icon]"></ng-content>
+            </div>
+            <div class="pxb-left-component">
+                <ng-content select="[left-component]"></ng-content>
+            </div>
+            <div class="mat-body-1 pxb-title" matLine [class.pxb-wrap]="wrapTitle">{{ title }}</div>
```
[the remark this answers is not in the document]
emclaug2: I think it would be best to leave any theme-related edits for the light/dark work.  I think the `mat-list-item` text is typically marked as a `matLine`.

components/src/core/info-list-item/info-list-item.component.ts
```diff
@@ -0,0 +1,58 @@
...
+            <div class="mat-body-2 pxb-subtitle" matLine [class.pxb-wrap]="wrapSubtitle">
+                <ng-container *ngIf="subtitleIsArray">
+                    <ng-container *ngFor="let sub of subtitle; let last = last">
+                        {{ sub }}<ng-container *ngIf="!last">{{ subtitleSeparator }}</ng-container>
```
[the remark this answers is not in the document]
emclaug2: Yep, logged a feature request here: https://github.com/pxblue/angular-component-library/issues/37

components/src/core/info-list-item/info-list-item.component.ts
```diff
@@ -0,0 +1,58 @@
...
+                    </ng-container>
+                </ng-container>
+                <ng-container *ngIf="!subtitleIsArray">{{ subtitle }}</ng-container>
+            </div>
+            <pxb-spacer flex="1" class="pxb-spacer"></pxb-spacer>
+            <div class="pxb-right-component">
+                <mat-icon *ngIf="chevron">chevron_right</mat-icon>
+                <ng-content *ngIf="!chevron" select="[right-component]"></ng-content>
```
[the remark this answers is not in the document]
emclaug2: I flipped the logic, chevron won't appear if the rightComponent is added.

docs/Spacer.md
emclaug2: Updated this & added a test.

docs/Spacer.md
emclaug2: Updated to 'Attributes'

docs/InfoListItem.md
emclaug2: You're correct, updated this.

components/src/core/utility/spacer.module.ts
```diff
@@ -0,0 +1,25 @@
+import { Component, HostBinding, Input, NgModule } from '@angular/core';
+
+@Component({
+    selector: 'pxb-spacer',
+    template: `
+        <ng-content></ng-content>
+    `,
+})
+export class SpacerComponent {
+    @Input() flex = 0;
```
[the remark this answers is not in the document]
emclaug2: Updated default to 1

components/src/core/info-list-item/info-list-item.component.ts
emclaug2: Good catch! I updated this.

emclaug2:
> * Can we update the stoybook full config to set a max width on the ILI to show the wrapping behavior via resizing instead of having to type in a bunch of text?
> * We also need to figure out what's going on in CircleCI...we're burning through 10 build minutes every time the storybook build times out.

The circle CI was failing because I was importing the Spacer module relatively from the components folder instead of from @pxblue/angular-components.  The build is fixed now.  

I'll update the Full config example size to be responsive.

emclaug2:
> Our storybook bundle is also getting quite large...we may want to start thinking about how we can optimize this, especially if we're planning to add a bunch of new components:
> image

This is a little concerning; I wonder if react has the same large bundle sizes.  This'll be a future enhancement.  I updated the webpack config to match react's in the meantime.

PR #38, opened 2020-04-10
Feature/comp tweaks

emclaug2:
- Hide InofListItem padding only when icon not present
- Remove @pxblue/colors dependency from @pxblue/angular-components
- Update scorecard action item styles to be 32x
- Update all components ChangeDetection strategy to be OnPush. 
- Update all components view encapsulation to be 'None'
- Prefix all components with the name `pxb-component-name-`.
- Update various unit tests
- Fix Empty State's margin issue
- Remove ChannelValue inline style Input because it conflicts with theming.
- Update to use the latest beta package of @pxblue/angular-themes

ChangeDetectionStrategy.OnPush reduces the amount of the ChangeDetection checks that'll happen within our components.  OnPush tells the UI to only redraw when an input to our component has changed.  This is appropriate because none of our components so far have any state-managing logic that would require a redraw.  The Inputs dictate what is on the screen.  This is the setting that angular material is using so I thought it'd be appropriate to use here as well.

ViewEncapsulation.None is angular material's component encapsulation setting as well.  Angular's default style encapsulation adds an invisible wrapper around each component so that styles won't be shared between components, despite duplicate class names.  Using ViewEncapsulation.None exposes our components classes to be targeted by 1. people using our components, or 2. Our global theme without the use of ::ng-deep.
::ng-deep is deprecated and might be removed in the future releases of Angular so it's best to follow suite with angular material.  

I also used angular material's class prefix style.

### DIFF
--- a/components/src/core/channel-value/channel-value.component.ts
+++ b/components/src/core/channel-value/channel-value.component.ts
@@ -1,15 +1,17 @@
-import { Component, Input } from '@angular/core';
+import {ChangeDetectionStrategy, Component, Input, ViewEncapsulation} from '@angular/core';
 
 @Component({
     selector: 'pxb-channel-value',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None,
     template: `
-        <span class="value-wrapper" [style.color]="color" [style.fontSize.px]="fontSize">
-            <span class="icon">
+        <span class="pxb-channel-value" [style.color]="color" [style.fontSize.px]="fontSize">
+            <span class="pxb-channel-value-icon">
                 <ng-content></ng-content>
             </span>
-            <h5 *ngIf="units && prefix" class="text units">{{ units }}</h5>
-            <h5 *ngIf="value" class="text value">{{ value }}</h5>
-            <h5 *ngIf="units && !prefix" class="text units">{{ units }}</h5>
+            <div *ngIf="units && prefix" class="pxb-channel-value-text pxb-channel-value-units">{{ units }}</div>
+            <div *ngIf="value" class="pxb-channel-value-text pxb-channel-value-value">{{ value }}</div>
+            <div *ngIf="units && !prefix" class="pxb-channel-value-text pxb-channel-value-units">{{ units }}</div>
         </span>
     `,
     styleUrls: ['./channel-value.component.scss'],
@@ -19,5 +21,5 @@ export class ChannelValueComponent {
     @Input() units: string;
     @Input() fontSize = 'inherit';
     @Input() prefix = false;
-    @Input() color = 'inherit';
+    @Input() color;
 }
--- a/components/src/core/channel-value/channel-value.component.ts
+++ b/components/src/core/channel-value/channel-value.component.ts
@@ -1,4 +1,4 @@
-import {ChangeDetectionStrategy, Component, Input, ViewEncapsulation} from '@angular/core';
+import { ChangeDetectionStrategy, Component, Input, ViewEncapsulation } from '@angular/core';
 
 @Component({
     selector: 'pxb-channel-value',
--- a/components/src/core/empty-state/empty-state.component.html
+++ b/components/src/core/empty-state/empty-state.component.html
@@ -1,9 +1,9 @@
-<div class="container" [style.color]="Colors.gray[500]">
-    <div style="line-height: 1">
+<div class="pxb-empty-state">
+    <div class="pxb-empty-state-empty-icon-wrapper">
         <ng-content select="[empty-icon]"></ng-content>
     </div>
-    <h2>{{ title }}</h2>
-    <h4 *ngIf="description" [style.color]="Colors.blue[500]">{{ description }}</h4>
+    <h2 class="pxb-empty-state-title">{{ title }}</h2>
+    <h4 *ngIf="description" class="pxb-empty-state-description">{{ description }}</h4>
     <div>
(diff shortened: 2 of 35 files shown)


components/src/core/score-card/scorecard.component.scss
```diff
@@ -42,6 +42,12 @@
     .pxb-action-items-wrapper {
         text-overflow: unset;
         overflow: unset;
+       ::ng-deep > * {
+            cursor: pointer;
+            height: 32px;
+            width: 32px;
+            text-align: center;
+        }
```
joebochill, line 50: It would appear we didn't actually make this click-target update in react (only react native). The spacing is off a little still. The final positioning is supposed to be:
- Right edge of the rightmost icon (24px) is 16px from the right edge of the card
- 12px between icons (from the right edge of the left icon viewbox to the left edge of the right icon viewbox)

In RN, we achieved the larger touch targets using negative margins:
- icon wrapper height/width 40px with 8px padding and -8px margin
- icons size 24px

I don't know if this is crucial for web components. If we want to match react, we can just do a 12px margin between the icons.

emclaug2: Ok I've updated the icons to be 24x24 and fixed the right padding on the 'action-items' container.  Now the card padding should be fine.

joebochill: did you push those changes?

emclaug2: Yep, just pushed them out.  With more changes, updating PR desc now.

emclaug2:
This needs to be published as a beta package before this PR can be merged. https://github.com/pxblue/themes/pull/65

joebochill:
> This needs to be published as a beta package before this PR can be merged. [pxblue/themes#65](https://github.com/pxblue/themes/pull/65)

which one needs to go first? This one or the theme?

components/src/core/channel-value/channel-value.component.ts
```diff
@@ -19,5 +21,5 @@ export class ChannelValueComponent {
     @Input() units: string;
     @Input() fontSize = 'inherit';
     @Input() prefix = false;
-    @Input() color = 'inherit';
+    @Input() color;
```
joebochill, line 24: do we need this at all if we're removing 'style' attributes and relying on SCSS?

components/src/core/empty-state/empty-state.component.html
```diff
@@ -1,9 +1,9 @@
-<div class="container" [style.color]="Colors.gray[500]">
-    <div style="line-height: 1">
+<div class="pxb-empty-state">
+    <div class="pxb-empty-state-empty-icon-wrapper">
```
joebochill, line 2: do we want to drop the 'wrapper' part of our class names?

components/src/core/empty-state/empty-state.component.spec.ts
```diff
@@ -20,77 +22,86 @@ import { EmptyStateComponent } from './empty-state.component';
...
+        const descriptionElement = fixture.debugElement.query(By.css('h4'));
+        expect(descriptionElement).toBeFalsy();
     });
 
     // Action Check
     it('should show actions when supplied', () => {
-        const fixture = TestBed.createComponent(TestEmpty);
-
-        let actionElement = fixture.debugElement.query(By.css('.withActions [actions]'));
+        const actikonFixture = TestBed.createComponent(TestEmpty);
```
joebochill, line 75: what's an actikon? ;)

components/src/core/hero/hero.component.scss
```diff
@@ -1,61 +1,59 @@
...
-    font-weight: 600;
-    width: 100%;
-    text-align: center;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    white-space: nowrap;
-    margin: 0;
-}
+    .pxb-hero-primary-wrapper {
+        padding: 3px;
```
joebochill, line 15: 4px to get us on the material spacing units?

components/src/core/empty-state/empty-state.component.html
emclaug2: I put the 'wrapper' part because it's wrapping a ng-content.  It's not the content itself so I think the wrapper part is appropriate.

For styling ng-content I've been following this is class naming convention: 
`ng-content-selector-name-wrapper`

components/src/core/empty-state/empty-state.component.spec.ts
emclaug2: Fixed!

components/src/core/hero/hero.component.scss
emclaug2: Good idea, updated to 4

components/src/core/channel-value/channel-value.component.ts
emclaug2: Huayun is working on edits that will remove this input.  I also think fontSize should go away, along with all respective stories/knobs.

emclaug2:
> > This needs to be published as a beta package before this PR can be merged. [pxblue/themes#65](https://github.com/pxblue/themes/pull/65)
> 
> which one needs to go first? This one or the theme?

I was thinking it makes sense to publish the themes first, then integrate them into this PR.

components/src/core/empty-state/empty-state.component.html
joebochill: sounds good to me, as long as we're consistent

PR #47, opened 2020-04-21
Feature/aot compat

### LINKED ISSUE #30: Component Library Not working with AOT-enabled projects

emclaug2:
Fixes #30 #31 

- Run Showcase with AOT to enforce AOT compatibility
- Explicitly export from `core/index` instead of `core` because ng-pkgr has issues with barrel files.
- MatDivider has to be imported from nested folder to work in NG9


emclaug2:
I looked at how angular material is doing their builds, and they are not using the ng-pkgr tool that is currently recommended by the official angular documentation.  They are using an experimental builder, bazel to create their nested exports.   I read the angular documentation and there is no mention of how to support nested exports (`@pxblue/angular-components/core/scorecard`).

emclaug2:
> I looked at how angular material is doing their builds, and they are not using the ng-pkgr tool that is currently included when generating a library through the angular CLI. They are using an experimental builder, bazel to create their nested exports. I read the angular documentation and there is no mention of how to support nested exports (`@pxblue/angular-components/core/scorecard`).

joebochill:
So we will do all of our imports from the root of the package, and this works fine for everything >= angular 7?

Do we need to update any of the docs for this?

emclaug2:
> So we will do all of our imports from the root of the package, and this works fine for everything >= angular 7?
> 
> Do we need to update any of the docs for this?

Yes now all of our imports will come from `@pxblue/angular-components`.

We used to support nested imports so I'll add a note to the changelog about the deep path deprecation, also there is a place or two in our current readmes that use a deep path in the example.   I'll update those references.

emclaug2:
> So we will do all of our imports from the root of the package, and this works fine for everything >= angular 7?
> 
> Do we need to update any of the docs for this?

Since our components only have peer dependencies on angular & angular-material, I'm thinking we should be fine in terms of compatibility.  

Showcase is using Angular v7.
Our library is built using all v8 depedencies.
I integrated our library in the angular-design-pattern POC which is using Angular v9.

So far so good.

PR #51, opened 2020-04-22
Feature/drawersubheader and drawerfooter

daileytj:
Fixes PXBLUE 1091.

- Add DrawerSubheaderComponent
- Add DrawerFooterComponent


emclaug2:
I'd be curious to see how these are used when we have the `<pxb-drawer>` component implemented.

Is the intention to use these components like this?
`<pxb-drawer>
   <pxb-drawer-header header>...</pxb-drawer-header>
   <pxb-drawer-subheader subheader>...</pxb-drawer-subheader>
   <pxb-drawer-footer footer>...</pxb-drawer-footer>
</pxb-drawer>`

I'm not sure if these needed to be individual components since I don't see a major difference between the above and this:

`<pxb-drawer>
   <pxb-drawer-header header>...</pxb-drawer-header>
   <div subheader>...</div>
   <div footer>...</div>
</pxb-drawer>`

I think it'll be more clear when the Drawer is implemented.

Issue #37, opened 2020-04-09
Enhance Subtitle Array in InfoListItem


emclaug2:
#### Describe the desired behavior
Current subtitle array only works with strings.

#### Describe the current behavior
Allow a user to pass in more than just a string array[].
Maybe also expand the [title] and [subtitle] to accept <ng-content>.

#### Is this request related to a current issue?
No

#### Additional Context
No

andrupu:
Should we let subtitle accept ng-content?

emclaug2:
Closing due to #52

PR #52, opened 2020-04-22
Feature/infolistitem ng content

### LINKED ISSUE #41: InfoListItem title should support ng content
InfoListItem title should support ng content

It only takes a string type input

No

emclaug2:
Fixes #41  #37 

- Make title in InfoListItem ng-content
- Make subtitle in InfoListItem ng-content
- Rename `rightComponent` and `leftComponent` to `rightContent` and `leftContent`


docs/ScoreCard.md
```diff
@@ -74,7 +74,7 @@ Child element with attributes:
 
 <div style="overflow: auto;">
 
-| Attributes  | Description                                 | Required | Default |
+| Selector    | Description                                 | Required | Default |
```
huayunh, line 77: May want to change everything

emclaug2: You're right, that was an accident change.  I think it makes sense to change `attributes` to `selector`, but not as a part of this PR.

emclaug2:
> So can we not pass in title and subtitle as strings anymore?

I think <ng-content> is a catch-all when it comes to these kind of things.  
If someone passes in a `<div title>Title Here</div>` or a span, our typography is still applied to the element they pass in.

If it's a convenience thing, I can add back the string inputs.  It's just kind of confusing to me to have two entry points to the same thing in the component.

joebochill:
> > So can we not pass in title and subtitle as strings anymore?
> 
> I think `<ng-content>` is a catch-all when it comes to these kind of things.
> If someone passes in a `<div title>Title Here</div>` or a span, our typography is still applied to the element they pass in.
> 
> If it's a convenience thing, I can add back the string inputs. It's just kind of confusing to me to have two entry points to the same thing in the component.

I think the markup for a user is cleaner with the attribute approach (since title will be used every time you use a ILI), but this is fine for now...we'll see if we get any pushback on it and we can add it back later if needed.

PR #53, opened 2020-04-23
Feature/list item tag

andrupu:
This pull request is for PXBLUE-1237

- Create a ListItemTag component built for Angular
--  The component(s) should be added to the angular-component-library repository
- Define the API for the component and verify with the team before starting development
--  This should match as closely as possible to what is available with the React component
- Implement the component
--  This only needs to look good in the light theme for now


emclaug2:
You need to resolve the merge conflicts with dev before this can be merged.

components/package.json
```diff
@@ -14,7 +14,7 @@
         "lint:fix": "eslint \"src/**/**.ts\" --fix"
     },
     "peerDependencies": {
-        "@angular/material": ">=7.0.0"
+        "@angular/material": ">=8.2.3"
```
emclaug2, line 17: This should not be changed.

components/src/core/list-item-tag/list-item-tag.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="pxb-list-item-tag-wrapper"
```
emclaug2, line 1: Root class should be `pxb-list-item-tag`.

components/src/core/list-item-tag/list-item-tag.component.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  beforeEach(() => {
+    fixture = TestBed.createComponent(ListItemTagComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
```
emclaug2, line 26: We need a test to enforce class names, see the other components for reference here.

components/src/core/list-item-tag/list-item-tag.component.ts
```diff
@@ -0,0 +1,19 @@
...
+  selector: 'pxb-list-item-tag',
+  templateUrl: './list-item-tag.component.html',
+  styleUrls: ['./list-item-tag.component.scss'],
+})
+
+export class ListItemTagComponent {
+
+  ngOnInit() {
+
+    console.log( "LOOK HERE " + this.fontColor);
```
emclaug2, line 13: Please lint your codebase, console logs aren't allowed in the code.

emclaug2: run `yarn lint` in the components folder.

components/src/core/list-item-tag/list-item-tag.component.ts
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  @Input() label: string;
```
emclaug2, line 16: @Inputs should be placed before the ngOnInit.

components/src/core/list-item-tag/list-item-tag.module.ts
```diff
@@ -0,0 +1,10 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { ListItemTagComponent } from './list-item-tag.component';
+
+@NgModule({
+    declarations: [ListItemTagComponent],
+    imports: [CommonModule],
```
emclaug2, line 7: Not sure if CommonModule is actually needed here.

demos/storybook/stories/list-item-tag/_module.stories.ts
```diff
@@ -0,0 +1,30 @@
...
+} from '../../src/constants';
+import { getReadMe, getReadMeStory, storyWrapper, UtilModule } from '../../src/utils';
+import { withBasicConfig } from './with-basic-config.stories';
+import { withFullConfig } from './with-full-config.stories';
+import { withinAnInfoListItem } from './within-an-InfoListItem.stories';
+
+storiesOf(`${COMPONENT_SECTION_NAME}/List Item Tag`, module)
+    .addDecorator(
+        moduleMetadata({
+            imports: [ListItemTagModule, MatButtonModule, MatIconModule, UtilModule, InfoListItemModule],
```
emclaug2, line 21: Why is MatButtonModule imported here?  This might not be needed.

demos/storybook/stories/list-item-tag/with-basic-config.stories.ts
```diff
@@ -0,0 +1,18 @@
+import { text } from '@storybook/addon-knobs';
+// import * from '@pxblue/angular-themes';
+import * as Colors from '@pxblue/colors';
+import { WITH_MIN_PROPS_STORY_NAME } from '../../src/constants';
+
+export const withBasicConfig = (): any => ({
+
+    name: WITH_MIN_PROPS_STORY_NAME,
```
emclaug2, line 8: Name not needed here, it's redefined in the `_module.stories.ts` file.

demos/storybook/stories/list-item-tag/within-an-InfoListItem.stories.ts
```diff
@@ -0,0 +1,19 @@
+import * as Colors from '@pxblue/colors';
+// import { InfoListItem, ListItemTag } from '@pxblue/angular-components';
```
emclaug2, line 2: Please lint the demos/storybook folder.

huayunh:
* Need to add it to the showcase, as the rightContent of `pxb-info-list-item`, see what React does. 

* In the readme, format all the "ListItemTag" as 
> ## **List Item Tag** 
> The **`<pxb-list-item-tag>`** is a text item with a...

  instead. Check other ng readmes.
* The "within an info list item" story is not showing the tag
image

* For the full config story, if a tag is labeled "clickable", then add an `action()` to it, or change the label to something else

components/src/core/list-item-tag/list-item-tag.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="pxb-list-item-tag-wrapper" 
+    
+    [style.backgroundColor]="backgroundColor">
```
joebochill, line 3: are we doing background and font color through attributes, or should we be leaving this to the user to customize through classes?

components/src/core/list-item-tag/list-item-tag.component.html
```diff
@@ -0,0 +1,9 @@
...
+    
+    <div class="mat-small" [class.pxb-list-item-tag-label]="label" [style.color]="fontColor">
+        {{ label.toUpperCase() }}
```
joebochill, line 6: How does a user set the style of the typography they want to use? (What's the equivalent to the `variant` prop from the react component?)

components/src/core/list-item-tag/list-item-tag.component.scss
```diff
@@ -0,0 +1,14 @@
...
+    
+    line-height: inherit;
+    padding-left: 4px;
+    border-radius: 2px;
+    padding-right: 4px;
+    letter-spacing: 1px;
+}
+
+.pxb-list-item-tag-label {
+    font-weight: bold;
```
joebochill, line 13: We don't typically use 'bold' as a value, we use the numerical weights (also we rarely use bold, most often we use semibold):
- 300 (light)
- 400 (regular)
- 600 (semibold)
- 700 (bold)
- 800 (extrabold)

docs/README.md
```diff
@@ -5,6 +5,6 @@ We currently have the following components available for Angular applications:
 - [Empty State](https://github.com/pxblue/angular-component-library/blob/dev/docs/EmptyState.md)
 - [Hero & Hero Banner](https://github.com/pxblue/angular-component-library/blob/dev/docs/Hero.md)
 - [Info List Item](https://github.com/pxblue/angular-component-library/blob/dev/docs/InfoListItem.md)
+- [ListItemTag](https://github.com/pxblue/angular-component-library/blob/dev/docs/ListItemTag.md)
```
joebochill, line 8: Please also update the changelog to include the addition of the listitemtag component

components/src/core/list-item-tag/list-item-tag.component.html
emclaug2: I think the bg/font color @inputs are fine in this case because the component is so small.   The component should be able to support styles through classes here if there is no @input colors provided.  This'll have to be true for the theme to work.

emclaug2: I dont have any issues removing the font/bg color inputs for the sake of consistency.

joebochill: I like the idea of consistency, but I fear that if we take them out, then the component doesn't really do anything. I'm okay with keeping them for the time being.

emclaug2: Sounds good to me, I think the component can exist without the input/bg color inputs if we do decide to remove them in the future.  It'll still have a few things going for it - PXBTheme, typography & padding so it looks great in our InfoListItem.

docs/README.md
andrupu: That should be in there.  I had included it in between InfoListItem and Scorecard

huayunh: Need space between the words, `List Item Tag`

demos/storybook/stories/list-item-tag/within-an-InfoListItem.stories.ts
andrupu: I was able to run prettier, but when kept getting an error when I ran yarn lint and yarn lint:fix.  May I have another set of eyes check this out.

andrupu:
> * Need to add it to the showcase, as the rightContent of `pxb-info-list-item`, see what React does.
> * In the readme, format all the "ListItemTag" as
> 
> > ## **List Item Tag**
> > The **`<pxb-list-item-tag>`** is a text item with a...
> 
> instead. Check other ng readmes.
> 
> * The "within an info list item" story is not showing the tag
>   image
> * For the full config story, if a tag is labeled "clickable", then add an `action()` to it, or change the label to something else

This was working when I initially Pull Requested it, but then I merged dev into this branch.  I took a look and can't pinpoint what went wrong.  Can I get another set of eyes on this.

emclaug2:
> > * Need to add it to the showcase, as the rightContent of `pxb-info-list-item`, see what React does.
> > * In the readme, format all the "ListItemTag" as
> > 
> > > ## **List Item Tag**
> > > The **`<pxb-list-item-tag>`** is a text item with a...
> > 
> > 
> > instead. Check other ng readmes.
> > 
> > * The "within an info list item" story is not showing the tag
> >   image
> > * For the full config story, if a tag is labeled "clickable", then add an `action()` to it, or change the label to something else
> 
> This was working when I initially Pull Requested it, but then I merged dev into this branch. I took a look and can't pinpoint what went wrong. Can I get another set of eyes on this.

`<pxb-info-list-item>
    <div title>Info List Item</div>
    <div subtitle>with a ListItemTag component to the right</div>
      ...
</pxb-info-list-item>`

The InfoListItem API has been changed.  Try something like the above.
Also the `right-component` selector is now `rightContent`.  This is why your tags aren't showing.

components/src/core/list-item-tag/list-item-tag.component.html
huayunh: react uses `variant=overline` by default. And unfortunately, Angular's obsolete typography does not have an overline class in their typography library, which also changes the capitalization with just CSS, and what you are seeing right here is apparently a custom style. 

So, I guess, at least use CSS property `text-transform: uppercase;` for this one

components/src/core/list-item-tag/list-item-tag.component.spec.ts
andrupu: Fixed with commit 8327066.

components/src/core/list-item-tag/list-item-tag.component.ts
andrupu: Thanks for the tip.

demos/storybook/stories/list-item-tag/within-an-InfoListItem.stories.ts
andrupu: Got it fixed with commit 6918cb5

emclaug2:
> * Need to add it to the showcase, as the rightContent of `pxb-info-list-item`, see what React does.
> * In the readme, format all the "ListItemTag" as
> 
> > ## **List Item Tag**
> > The **`<pxb-list-item-tag>`** is a text item with a...
> 
> instead. Check other ng readmes.
> 
> * The "within an info list item" story is not showing the tag
>   image
> * For the full config story, if a tag is labeled "clickable", then add an `action()` to it, or change the label to something else

The full config story name was changed. 
The documentation was updated to be aligned with the others. 
The Showcase was updated to include the tag. 
The contextual story was updated to show the tag.

PR #58, opened 2020-04-24
Bug/infolistitem light dark sb

### LINKED ISSUE #54: InfoListItem Storybook Needs Light/Dark Support
#### Describe the bug
InfoListItem in storybook does not respond well to the dark theme.

Background should not be white.

Go to storybook.
Open a lnfoListItem story.
Click the dark toggle.
Observe the white background in the InfoListItem.

 Framework
- [ x] Angular
- [ ] React
- [ ] Ionic
- [ ] React Native

 Browser
- [ ] Internet Explorer
- [ ] Edge
- [ ] Safari
- [x ] Chrome
- [ ] Firefox

 Operating System
- [ ] MacOS
- [x ] Windows
- [ ] Other

emclaug2:
Fixes #54 

- Add light dark to the InfoListItem examples

### DIFF
--- a/demos/storybook/stories/info-list-item/_module.stories.ts
+++ b/demos/storybook/stories/info-list-item/_module.stories.ts
@@ -7,7 +7,7 @@ import {
     STORY_PARAMS, WITH_FULL_CONFIG_STORY_NAME,
     WITH_MIN_PROPS_STORY_NAME,
 } from '../../src/constants';
-import { getReadMe, getReadMeStory, storyWrapper, UtilModule } from '../../src/utils';
+import {getReadMe, getReadMeStory, isDarkMode, storyWrapper, UtilModule} from '../../src/utils';
 import { withBasicConfig } from './with-basic-config.stories';
 import { withIcon } from "./with-icon.stories";
 import { withSubtitle } from "./with-subtitle.stories";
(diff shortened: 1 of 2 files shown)


huayunh:
It looks like react has similar excitements there. Logging issues now... 

image

Consider changing the icon color / avatar color to something else?

huayunh:
Or maybe change the default color 

image

React is using gray300

emclaug2:
Using the default color provided by the themes is a good idea; I took out the manual styling (it was Black[500] before for light/dark).  I also updated the black[500] avatar color so it stands out in both light/dark themes. 

Here's the new dark theme: 
image

emclaug2:
Here's the light theme too: 
image

huayunh:
So the dark info list item background is not part of the default component behavior? 

Other than that, everything looks good to me. Besides my fetish to prettier everything.

emclaug2:
> So the dark info list item background is not part of the default component behavior?
> 
> Other than that, everything looks good to me. Besides my fetish to prettier everything.

Right now the React and Angular InfoListItem backgroundColor is inherit, so it needs to be wrapped in a div that'll update when the light/dark theme is applied.

If the InfoListItem is supposed to have a color then we'd have to bake this change into the component and not the storybook.

huayunh:
I see, that explains. Thanks!

huayunh:
Looks like the foreground color is not inheriting yet. I will log a separate issue on this

PR #73, opened 2020-05-19
Accessibility update

huayunh:
Turns out Angular already has some accessibility issues pre-addressed. We may not even need to update anything in themes for Accessibility purposes. 

- Updated a bunch of colors in the storybook
- Added prettier back
- Margin tweaks on empty state


emclaug2:
> Can you update the ChannelValue 'with icon' story use a darker icon color? It's still throwing an accessibility error.

Looking closer, this story has a knob for icon color.  I think the default color should be a a11y-compliant one.

huayunh:
@emclaug2 icons get recognized as a font here, but in reality, it should be a "UI element", which only has an accessibility contrast ratio requirement of 3:1.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change Log
 
+## v2.0.1
+-  Twists styles to address some accessibility concerns.
```
joebochill, line 4:
```suggestion
-  Colors updated for accessibility
```

PR #77, opened 2020-06-03
Display ver number for sb

huayunh:
Fixes PXBLUE-1144

- Add to link scripts an additional step to copy the package.json from `/components` to `demos/storybook/assets` as `_package.json`
- have the welcome page import and read from `_package.json`'s `version` field, and have it displayed under the "Angular Components Library" title

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,7 @@ jobs:
           name: Build Storybook
           command: |
             cp -r ./dist/. ./demos/storybook/node_modules/@pxblue/angular-components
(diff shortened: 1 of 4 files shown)


huayunh:
Ooof, it worked locally. Let me see...

joebochill:
What about just reading the package.json from node modules?
```
import pjson from ‘@pxblue/angular-component-library/package.json’
```

That way it will always be up to date with the version being used and you won’t have to worry about the file not existing prior to link:components running (ie you could run the storybook as-is after a fresh clone...or via a code sandbox link)

huayunh:
That is brilliant!

PR #80, opened 2020-06-09
Feature/drawer refactor

emclaug2:
- Refactor Drawer
- Add animations on open/close of drawer
- Fix DrawerNavGroup items overlap when animating
- Add support for nested, nested items
- Add drawer demo to showcase
- Add drawer stories to storybook

InfoListItem changes:
- dense width decreased from 56px to 52px (matches react)

Can be ran locally or seen here: https://pxblue-angular-library.web.app


emclaug2:
Todo:
- Expansion panel nested styles (removed round corners).  
- Add support for custom expand/collapse icons

emclaug2:
> Been messing around with the transition jankiness for an hour now with no progress. If there's anything else that needs addressed I'd much rather do that.

Spend a few hours on this with no luck; this is a jankiness is a bigger time sink than expected.

joebochill:
I'm okay with the jankiness for now (let's log an issue for it) since it's only a problem when you go between variants.

For the stories:
- What about a content-type switcher for the 'with subheader' (react has a radio for using a search box or an accordion)? Looks like it also has an Input for `divider`
- Are there no knobs/inputs available for the 'with nested nav items' story (I see divider as an Input for nav group and nav item)?
- For 'with footer' footer has Input for `divider`

What was the reasoning for moving the `variant` to the drawer-layout? Is it because the drawer-layout needs to know what kind of drawer we have to properly resize the content? How does Angular Material manage to do it on the <drawer>?

emclaug2:
> I'm okay with the jankiness for now (let's log an issue for it) since it's only a problem when you go between variants.
> 
> For the stories:
> 
> * What about a content-type switcher for the 'with subheader' (react has a radio for using a search box or an accordion)? Looks like it also has an Input for `divider`
> * Are there no knobs/inputs available for the 'with nested nav items' story (I see `divider` as an Input for nav group and nav item)?
> * For 'with footer' footer has Input for `divider`
> 
> What was the reasoning for moving the `variant` to the drawer-layout? Is it because the drawer-layout needs to know what kind of drawer we have to properly resize the content? How does Angular Material manage to do it on the `<drawer>`?

I'll gladly write up a new issue to fix the jankiness later; it would be great if we could get this fixed before R16 release just so we don't hear about it from users. 

I can add a content switcher/divider knob for the subheader.
I can add knobs for the nested nav group story to make it more interesting.
I'll add two knobs on the footer story as well.

The reason why the `variant` was moved to the DrawerLayout instead of the Drawer is because our Drawer and the `mat-sidenav` don't really use the variant.  Angular material's container component `mat-sidenav-container` is what needs to use the variant to determine the `mode` of how things are arranged.

emclaug2:
> What was the reasoning for moving the `variant` to the drawer-layout? Is it because the drawer-layout needs to know what kind of drawer we have to properly resize the content? How does Angular Material manage to do it on the `<drawer>`?

Currently the drawer is not a wrapper of the mat-sidenav.  If we make the top-level component of the Drawer mat-sidenav then it will make more sense.

Having the variant on the drawer makes me think that a `variant` knob should work when the drawer is outside of a DrawerLayout.  This will require a little bit of refactor work just to see if it'll work.

emclaug2:
>  (I see `divider` as an Input for nav group and nav item)?

Yes the NavGroup divider adds a divider under a nav group title.  The NavItem divider is inbetween InfoListItems.  There is no inheritance in this implementation; most props will be placed directly on the pxb-nav-item.

joebochill:
> Having the variant on the drawer makes me think that a `variant` knob should work when the drawer is outside of a DrawerLayout. This will require a little bit of refactor work just to see if it'll work.

If you're looking to spend some time i'd be interested to see how that plays out. The angular material drawer can work outside of a sidenav-container (if you wanted to use a drawer in some other component i guess), so i would expect that ours could too. If we can't make it work for now, I'm fine with how we have it for this release and we can circle back down the road. More often than not, people are going to only be using it in the drawerlayout anyway.

joebochill:
Logged #83 to capture the refactor work for the variant location

PR #87, opened 2020-06-22
Move host selector styles to component parent class

### LINKED ISSUE #44: ChannelValue not taking :host styles
#### Describe the bug
In channel-value/channel-value.component.scss there is this field
```
:host {
    display: inline-flex;
    align-items: center;
}
```
but this style applied neither on storybook nor on showcase. 

Styles should go through so that the component `pxb-channel-value` can be properly vertically aligned in an info list item. This is also what React Component Library did in ChannelValue's React Counter Part. 

1. Go to showcase
2. See the list of info list items

In showcase,
image

 Browser
- [ ] Internet Explorer
- [ ] Edge
- [ ] Safari
- [x] Chrome
- [ ] Firefox

 Operating System
- [x] MacOS
- [ ] Windows
- [ ] Other

daileytj:
Fixes #44.

- Move host selector styles to the component parent class

### DIFF
--- a/components/src/core/info-list-item/info-list-item.component.scss
+++ b/components/src/core/info-list-item/info-list-item.component.scss
@@ -109,8 +109,9 @@ $denseHeight: 52px;
         order: 5;
         margin-left: 16px;
         white-space: nowrap;
-        .pxb-info-list-item-right-content-wrapper * {
+        .pxb-info-list-item-right-content-wrapper > * {
             display: flex;
+            align-items: center;
         }
     }
 }
--- a/components/src/core/info-list-item/info-list-item.component.scss
+++ b/components/src/core/info-list-item/info-list-item.component.scss
@@ -89,8 +89,8 @@ $denseHeight: 52px;
 
     .pxb-info-list-item-left-content-wrapper {
         order: 2;
-        margin-right: 16px;
-        * {
+        > * {
+            margin-right: 16px;
             display: flex;
         }
     }
--- a/components/src/core/info-list-item/info-list-item.component.scss
+++ b/components/src/core/info-list-item/info-list-item.component.scss
@@ -89,8 +89,8 @@ $denseHeight: 52px;
 
     .pxb-info-list-item-left-content-wrapper {
         order: 2;
-        > * {
-            margin-right: 16px;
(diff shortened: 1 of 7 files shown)


daileytj:
If anyone is willing to investigate... it looks like there is still a margin/padding issue with the scorecard in the showcase.

demos/showcase/src/app/app.component.html
```diff
@@ -134,7 +134,7 @@ <h2>Showcase</h2>
                 </pxb-info-list-item>
 
                 <pxb-info-list-item divider="full" [subtitle]="['Phase A', 'Phase B', 'Phase C']" [avatar]="true">
-                    <div title>Input Voltage</div>
+                    <div pxb-title>Input Voltage</div>
```
joebochill, line 137: nice catch, I actually noticed this in another branch and was going to request someone fix it

huayunh:
So the idea is moving everything from `<pxb-channel-value>` tag to `.pxb-channel-value` class?

components/src/core/info-list-item/info-list-item.component.scss
```diff
@@ -89,8 +89,8 @@ $denseHeight: 52px;
 
     .pxb-info-list-item-left-content-wrapper {
         order: 2;
-        > * {
-            margin-right: 16px;
+        margin-right: 16px;
```
joebochill, line 92: This is the change that is causing the misalignment in the scorecard. We shouldn't set the right margin on the wrapper, we should be setting the right margin on each of the children (otherwise the margin will be there even if there is no content, which is the case in the showcase demo). Does this rule need to be changed? What was the issue you saw that prompted the modification?

components/src/core/score-card/score-card.component.scss
```diff
@@ -3,7 +3,7 @@
 }
 
 // Styles applied to supplied action items.pxb-score-card-
-:host ::ng-deep .pxb-score-card-action-items * {
+.pxb-score-card-action-items * {
```
joebochill, line 6: We might want to change this to be a little more restrictive in what it applies to (just direct children):
```suggestion
.pxb-score-card-action-items > * {
```

PR #97, opened 2020-06-25
Add third line of text to InfoListItem

### LINKED ISSUE #96: InfoListItem should support third line of text
**Describe the desired behavior**
Material Design Lists support up to three lines of text in list items

**Describe the current behavior**
InfoListItem only allows configuration for title and subtitle (2 lines)

**Is this request related to a current issue?**
No

daileytj:
Fixes #96.

- Add third line of text to InfoListItem.
- Update full config story to use "info" projected content

note: I didn't update the changelog since we haven't officially released the InfoListItem.

### DIFF
--- a/components/src/core/info-list-item/info-list-item.component.ts
+++ b/components/src/core/info-list-item/info-list-item.component.ts
@@ -47,6 +47,13 @@ import { requireContent, isEmptyView } from '../../utils/utils';
             >
                 <ng-content select="[pxb-subtitle]"></ng-content>
             </div>
+            <div
+                class="mat-body-2 pxb-info-list-item-info-wrapper"
+                matLine
+                [class.pxb-info-list-item-wrap]="wrapSubtitle"
+            >
+                <ng-content select="[pxb-info]"></ng-content>
+            </div>
             <pxb-spacer class="pxb-info-list-item-spacer"></pxb-spacer>
             <div class="pxb-info-list-item-right-content">
                 <div #right class="pxb-info-list-item-right-content-wrapper">
--- a/demos/storybook/stories/info-list-item/with-full-config.stories.ts
+++ b/demos/storybook/stories/info-list-item/with-full-config.stories.ts
@@ -18,6 +18,7 @@ export const withFullConfig = (): any => ({
             (click)="action()">
             <div pxb-title>{{title}}</div>
             <div pxb-subtitle>{{subtitle}}</div>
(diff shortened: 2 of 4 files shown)


emclaug2:
Here's a picture of how it looks for anyone who wants to see it: 

image

components/src/core/info-list-item/info-list-item.component.ts
```diff
@@ -47,6 +47,13 @@ import { requireContent, isEmptyView } from '../../utils/utils';
             >
                 <ng-content select="[pxb-subtitle]"></ng-content>
             </div>
+            <div
+                class="mat-body-2 pxb-info-list-item-info-wrapper"
+                matLine
+                [class.pxb-info-list-item-wrap]="wrapSubtitle"
```
emclaug2, line 53: This is the only thing that kinda stands out to me; do you think there would be benefit in introducing a new variable `wrapInfo`?  I really doubt this would get much use, but it does help with the flexibility of the component.

daileytj: I just used the wrapSubtitle, making the assumption that if you wanted the subtitle text to wrap then all of it should wrap(including the third line). We can certainly make that update if that's what we want to do. @joebochill and I were discussing the possibility of just having a wrapText prop that would make it an all or nothing kind of thing. Not sure which direction we want to go.

joebochill: We're also likely going to refactor this in an upcoming major release to be more like angular (take out the logic for doing the separator stuff and just allow any ReactNode as the subtitle).

emclaug2: Okay that's fine with me, I don't think it's an issue.  This PR looks good to me!

demos/storybook/stories/info-list-item/with-full-config.stories.ts
```diff
@@ -18,13 +18,15 @@ export const withFullConfig = (): any => ({
             (click)="action()">
             <div pxb-title>{{title}}</div>
             <div pxb-subtitle>{{subtitle}}</div>
+            <div pxb-info>{{info}}</div>
```
joebochill, line 21: So in Angular we added this knob in the full-config story and in React it's has its own story...we shuld be consistent.  I kind of like it in the 'basic' story, but if we'd rather try to keep people to just two lines in typical usage, then full-config probably makes more sense.

PR #101, opened 2020-06-26
Bug/sb drawer expansion panel

emclaug2:
- Fix bug where the expansion panel in the Drawer `with subheader` story wouldn't expand.

### DIFF
--- a/demos/storybook/stories/drawer/_module.stories.ts
+++ b/demos/storybook/stories/drawer/_module.stories.ts
@@ -26,6 +26,7 @@ import { withFullConfig } from './with-full-config.stories';
 import { withinDrawerLayout } from './within-drawer-layout.stories';
 import { MatToolbarModule } from '@angular/material/toolbar';
 import { MatExpansionModule } from '@angular/material/expansion';
+import {MatCardModule} from "@angular/material/card";
 
 export const drawerWrapper = () => (storyFn): any => {
     const story = storyFn();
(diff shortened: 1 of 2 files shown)


jeffvg:
I see the same behavior clicking accordion. Toggle only happens after clicking and moving pointer of click area.

ng-drawer-subheader

emclaug2:
You're right @jeffvg .  On Friday the accordion was working just fine for me but I see what you're saying; it's unresponsive until the mouse is moved outside of the container.  I'll take another look at this later in the day; definitely some strange behavior here.

huayunh:
That accordion subheader looks like a card from Jeff's screenshot...

emclaug2:
A little more information about this bug.   I think storybook is having some change detection issues when a component is conditionally rendered.  If you open the subheader demo with the accordion rendered by default, it'll work as expanded.  Open/close on click.  

https://pxblue-components.github.io/angular-dev/?path=/story/components-drawer--with-subheader&knob-divider=true&knob-Subheader Content=Accordion

If you toggle to `filter` then back to `accordion` using the knobs the open/close functionality is broken again.

emclaug2:
> That accordion subheader looks like a card from Jeff's screenshot...

I've replaced the accordion with a card to make things simpler.

emclaug2:
Forgot to Prettier & Lint! Should be good to go now.

PR #109, opened 2020-07-10
Feature/lint updates

emclaug2:
- Lint using the latest eslint-config rules


components/package.json
```diff
@@ -31,16 +31,16 @@
         "@angular/material": "^8.2.3",
         "@angular/platform-browser": "~8.2.2",
         "@angular/platform-browser-dynamic": "~8.2.2",
-        "@pxblue/eslint-config": "^1.1.4",
+        "@pxblue/eslint-config": "^2.0.1-beta.1",
```
joebochill, line 34: Let's avoid using beta packages in the dependency arrays for our packages when we can. If this is ready for a full publish, i'll promote it today.

components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.ts
```diff
@@ -105,14 +105,14 @@ export class DrawerNavItemComponent extends StateListener implements Omit<Drawer
...
-    isEmpty = (el): boolean => isEmptyView(el);
+    isEmpty = (el: ElementRef): boolean => isEmptyView(el);
     isNestedItem: boolean;
     drawerOpen: boolean;
     hasChildren = false;
     depth: number;
     id: number;
 
-    constructor(drawerService: DrawerService, private readonly changeDetectorRef: ChangeDetectorRef) {
+    constructor(drawerService: DrawerService, changeDetectorRef: ChangeDetectorRef) {
```
joebochill, line 115: are we changing these from private->public to avoid a linting rule, or is it because they make more sense to be public?

components/src/core/drawer/drawer.component.ts
```diff
@@ -71,7 +71,7 @@ export class DrawerComponent extends StateListener implements OnInit, OnChanges
 
     // Close drawer on selection if drawer is only temporarily opened.
     // Expanding nested navitems when temporarily opened will not close the drawer.
-    private listenForDrawerSelection(): void {
+    listenForDrawerSelection(): void {
```
joebochill, line 74: same question as above about private->public

components/src/utils/test-utils.ts
```diff
@@ -3,8 +3,9 @@ import { ComponentFixture } from '@angular/core/testing';
 // Using a selector, counts the number of instances.
 export const count = (fixture: ComponentFixture<any>, selector: string, expected = 1): void => {
     const instances = fixture.nativeElement.querySelectorAll(selector);
-    expect(instances.length).toBe(
+    const length: string = instances.length;
+    void expect(instances.length).toBe(
```
joebochill, line 7: expect (length) ?

emclaug2: Good catch, updated this.

components/src/core/drawer/drawer.component.ts
emclaug2: I'm not sure why this method was marked as private in the first place so I removed this modifier.  A lot of our components could potentially have some private methods on it but our default for now has been to make everything public, so I just kept with that approach for now.

components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.ts
emclaug2: Each Drawer subsection component inherits a `drawerService` and `changeDetector` from it's base class.  I was messing around with the access modifiers but finally settled on marking these services as `protected`.

components/package.json
emclaug2: I haven't used this package to test any react code yet, but from an angular perspective it's good to be published.

PR #115, opened 2020-07-21
Feature/user menu

emclaug2:
- UserMenu component added to core components
- Storybook, Showcase, and Docs updated to match React.

At this point, it would be nice to get an extra set of eyes on the API to make sure it makes sense and visually looks like things are being arranged in the correct place.

There are few things left TODO that I'll address tomorrow: 

- This is our first component in angular that uses overlays; overlays are generated outside the scope of an application's root node, so the UserMenu menu is outside the scope of our light/dark theme.  This explains why the InfoListItem's text color is so dark.
- Animations for Overlays have to manually added; right now there's no transition for when the menu is opened/dismissed. 

#### Screenshots:
image


emclaug2:
In the docs I have two ways to create the body of the UserMenu - by passing in a menuGroup object, or by projecting content.  The first way was inspired by how we handled this in react, but I think I should remove it and just stick with the latter approach.  This will be aligned with how we handle populating the DrawerBody, and it'll make our API more opinionated on how to use it.

joebochill:
> * This is our first component in angular that uses overlays; overlays are generated outside the scope of an application's root node, so the UserMenu menu is outside the scope of our light/dark theme.  This explains why the InfoListItem's text color is so dark.

Our instructions for applying the theme have a note about reminding people to also put the theme class on the global overlay container as well: https://pxblue.github.io/development/frameworks-web/angular#applying-the-theme

emclaug2:
> > * This is our first component in angular that uses overlays; overlays are generated outside the scope of an application's root node, so the UserMenu menu is outside the scope of our light/dark theme.  This explains why the InfoListItem's text color is so dark.
> 
> Our instructions for applying the theme have a note about reminding people to also put the theme class on the global overlay container as well: https://pxblue.github.io/development/frameworks-web/angular#applying-the-theme
> 

Thanks, Joe!  This did the trick.  (theme updates applied in pic)
image

Applying the theme to the entire body instead of just the app container fixes this problem.

joebochill:
> Applying the theme to the entire body instead of just the app container fixes this problem.

Can you write a story to make that the standard recommendation/practice (include updates to doc it instructions, themes readme, CLI integration function)?

docs/UserMenu.md
```diff
@@ -0,0 +1,107 @@
...
+The following child elements are projected into `<pxb-user-menu>`:
+
+<div style="overflow: auto;">
+
+| Selector           | Description                                                                    | Required | 
+| ------------------ | ------------------------------------------------------------------------------ | -------- | 
+| [pxb-avatar]       | Custom avatar to show, alternative to `avatarImage` or `avatarValue` prop  | no       | 
+| [pxb-menu-avatar]  | Custom menu avatar to show                                                     | no       |
+| [pxb-header]       | Custom menu header content                                                     | no       |
+| [pxb-body]         | Custom menu body content                                                       | no       |
```
joebochill, line 93: do we use pxb-body elsewhere? would pxb-menu-body be more descriptive?

components/src/core/user-menu/user-menu.component.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+            imports: [UserMenuModule],
+        }).compileComponents();
+        fixture = TestBed.createComponent(UserMenuComponent);
+        component = fixture.componentInstance;
+    });
+
+    it('should initialize', () => {
+        fixture.detectChanges();
+        void expect(component).toBeTruthy();
+    });
```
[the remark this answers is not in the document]
emclaug2: I added some more tests, I forgot to come back and add them.

docs/UserMenu.md
emclaug2: I updated pxb-body and pxb-header to be pxb-menu-header and pxb-menu-body.

docs/UserMenu.md
```diff
@@ -0,0 +1,107 @@
...
+## API
+
+Parent element (`<pxb-user-menu>`) attributes:
+
+<div style="overflow: auto;">
+
+| @Input       | Description                                      | Type                    | Required | Default |
+| ------------ | ------------------------------------------------ | ----------------------- | -------- | ------- |
+| avatarImage  | Image source for avatar                          | `string`                | no       |         |
+| avatarValue  | Text value for avatar                            | `string`                | no       |         |
```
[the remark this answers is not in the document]
emclaug2:
```
<div class="pxb-user-menu-avatar" [class.non-text-avatar]="!avatarValue">
            <ng-container *ngIf="avatarValue">{{ avatarValue }}</ng-container>
            <img *ngIf="avatarImage" [src]="avatarImage" alt="User Menu Avatar" />
            <ng-content></ng-content>
        </div>
```

It will try to render both inside the small avatar window.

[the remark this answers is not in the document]
emclaug2: Same goes for if they also provide a custom avatar content too, then it'll try to render all 3 at once.

components/src/core/user-menu/user-menu.component.scss
```diff
@@ -0,0 +1,37 @@
...
+mat-card.pxb-user-menu-overlay {
+    padding: 0;
+}
+
+.pxb-user-menu-header-avatar {
+    margin-left: 4px;
+}
+
+.pxb-user-menu-header mat-toolbar {
+    border-radius: 4px !important;
```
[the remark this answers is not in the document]
emclaug2: Good catch, updated this.

docs/UserMenu.md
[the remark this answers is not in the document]
joebochill: At some point we may want to think about going in and having a preference on things like this (e.g., if custom avatar, render than and nothing else, if no custom avatar, check if theres a image avatar, if not then look for value) so that a user doesn't accidentally run into the situation where they've got 3 avatars controlling the menu. On quick glance it doesn't look like we're doing this anywhere else either (Drawer Header, Navgroup header, etc.), so it should be alright for now.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## v2.1.0
+-   Add new component for `<pxb-user-menu` (requires @pxblue/angular-themes@^5.1.0 for theme styling).
```
joebochill, line 4:
```suggestion
-   Add new component for `<pxb-user-menu` (use @pxblue/angular-themes v5.1.0+ to get PX Blue themes for this component).
```

components/src/core/user-menu/user-menu.component.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+        void expect(header).toBeTruthy();
+    });
+
+    it('should render the nav items', () => {
+        const listFixture = TestBed.createComponent(UserMenuNavItemsTest);
+        listFixture.detectChanges();
+        const listItem = document.getElementById('list-item-test');
+        void expect(listItem).toBeTruthy();
+    });
+});
```
joebochill, line 98: Nice job with the tests

components/src/core/user-menu/user-menu.component.ts
```diff
@@ -0,0 +1,91 @@
...
+    @Input() avatarValue: string;
+    @Input() avatarImage: string;
+    @Input() menuTitle: string;
+    @Input() menuSubtitle: string;
+    @Input() open = false;
+
+    @Output() openChange: EventEmitter<boolean> = new EventEmitter<boolean>();
+    @Output() backdropClick: EventEmitter<void> = new EventEmitter<void>();
+
+    positions = [
```
joebochill, line 77: Can this position be changed by the end user if they want to customize where the menu opens relative to the avatar?

demos/storybook/stories/user-menu/with-custom-menu-header.stories.ts
```diff
@@ -0,0 +1,70 @@
+const Trex = require('../../assets/trex.png');
+const eatonLogo = require('../../assets/EatonLogo.svg');
+
+export const withCustomMenuHeader = (): any => ({
```
joebochill, line 4: The react SB calls this "with custom menu" and uses other components for the list items (instead of info-list-item)

demos/storybook/stories/user-menu/with-full-config.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+        title: 'Contact Us',
+        icon: 'mail',
+    },
+    {
+        title: 'Log Out',
+        icon: 'logout',
+    },
+];
+
+export const withFullConfig = (): any => ({
```
joebochill, line 18: This demo is identical to the 'with a menu header' story (see comment there). React has the avatar and menu props slit into two sections, but since there's only three on the angular side that is probably overkill.

demos/storybook/stories/user-menu/with-menu-header.stories.ts
```diff
@@ -0,0 +1,26 @@
...
+                <pxb-info-list-item *ngFor="let item of items" [dense]="true" (click)="open=false">
+                    <mat-icon pxb-icon>{{item.icon}}</mat-icon>
+                    <div pxb-title>{{item.title}}</div>
+                </pxb-info-list-item>
+            </mat-nav-list>
+    </pxb-user-menu> 
+    `,
+    props: {
+        open: false,
+        avatarValue: text('avatarValue', 'AV'),
```
joebochill, line 21: This story is the same as the full config. We can probably drop the avatarLabel knob here and just focus on the text part of the menu header (react only has the text knobs).

docs/UserMenu.md
```diff
@@ -0,0 +1,107 @@
...
+The `<pxb-user-menu>` is an Avatar that opens a Menu when clicked. It is typically used in the top-right corner of an application and indicates who is logged in.
+
+<div style="align-items: center; display:flex; justify-content: space-around">
+
+<img width="30%" alt="UserMenu Avatar" src="./images/userMenuAvatar.png">
+<img width="35%" alt="UserMenu Opened" src="./images/userMenuOpened.png">
+
+</div>
+
+The Menu can be populated via the `menuGroups` prop, or can be entirely customized by supplying your own content.  See the API section below for more details.
```
joebochill, line 12: Do we need to tweak the description up here? I didn't see a menuGroups prop on the Angular user-menu.

docs/UserMenu.md
```diff
@@ -0,0 +1,107 @@
...
+</pxb-user-menu> 
+```
+
+## API
+
+Parent element (`<pxb-user-menu>`) attributes:
+
+<div style="overflow: auto;">
+
+| @Input       | Description                                      | Type                    | Required | Default |
```
joebochill, line 66: If none of these inputs are required, what happens if none of them are supplied?

components/src/core/user-menu/user-menu.component.ts
emclaug2: Yes we can make this positions be an @Input so that a user can change the position of where the Menu will appear.

docs/UserMenu.md
emclaug2: Yes this was leftover from when we still created the menu body from the menuGroups prop.  I've updated it.

demos/storybook/stories/user-menu/with-menu-header.stories.ts
emclaug2: Good catch, I've reverted the story name and created the menu items using mat-list-items instead of info-list-items to showcase the difference.

docs/UserMenu.md
emclaug2: The avatar and menu will render, but nothing will be shown in the avatar container. 

image

emclaug2: image

demos/storybook/stories/user-menu/with-full-config.stories.ts
emclaug2: If I make the positions an @Input then then these new knobs should slightly beef up this story to be more exciting.

demos/storybook/stories/user-menu/with-custom-menu-header.stories.ts
emclaug2: Good catch, I've reverted the story name and created the menu items using mat-list-items instead of info-list-items to showcase the difference.

docs/UserMenu.md
joebochill: A little goofy, but I'll take a blue circle over nothing 👍

demos/storybook/stories/user-menu/with-menu-header.stories.ts
emclaug2: I dropped the avatarLabel here and made the full config story more exciting.

docs/UserMenu.md
```diff
@@ -0,0 +1,108 @@
...
+The `<pxb-user-menu>` is an Avatar that opens a Menu when clicked. It is typically used in the top-right corner of an application and indicates who is logged in.
+
+<div style="align-items: center; display:flex; justify-content: space-around">
+
+<img width="30%" alt="UserMenu Avatar" src="./images/userMenuAvatar.png">
+<img width="35%" alt="UserMenu Opened" src="./images/userMenuOpened.png">
+
+</div>
+
+The Menu header can be constructed by supplying a `menuTitle` and optional `menuSubtitle`, or your own custom content. The Menu body is entirely custom content.  See the API section below for more details.
```
joebochill, line 12:
```suggestion
The Menu header can be constructed by supplying a `menuTitle` and optional `menuSubtitle`, or your own custom content. The Menu body is entirely custom content. See the API section below for more details.  
```

joebochill: extra space in here between the last two sentences. 🤒

docs/UserMenu.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+## API
+
+Parent element (`<pxb-user-menu>`) attributes:
+
+<div style="overflow: auto;">
+
+| @Input       | Description                                      | Type                        | Required | Default                                                                                                                     |
+| ------------ | ------------------------------------------------ | --------------------------- | -------- | -------                                                                                                                     |
```
joebochill, line 67: Can we clean up these tables a little bit to make the column separators align? I wish there was a VSCode plugin for this

scripts/buildTest.sh
```diff
@@ -39,6 +39,10 @@ echo "Checking Drawer..."
 echo -ne "  drawer.component: "
 if [ ! -f ./core/drawer/drawer.component.d.ts  ]; then echo -e "${RED}Not Found${NC}" && exit 1; else echo -e "${GREEN}Found${NC}"; fi
 
+echo "Checking UserMenu..."
+echo -ne "  user-menu.component: "
+if [ ! -f ./core/user-menu/user-menu.component.d.ts  ]; then echo -e "${RED}Not Found${NC}" && exit 1; else echo -e "${GREEN}Found${NC}"; fi
+
```
joebochill, line 45: Can you also add checks for ILI, ListItemTag, and Spacer? We missed those when we implemented them.

emclaug2: Yep good catch, just added those in.

docs/UserMenu.md
emclaug2: Should be all  cleaned up now

docs/UserMenu.md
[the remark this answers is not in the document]
huayunh: That's something for docit design overhaul, probably not for our api

docs/UserMenu.md
huayunh: > I wish there was a VSCode plugin for this

If we `prettier` the UserMenu.md, I think it automatically does it for you. I have our prettier hooked up with VSC on save to satisfy my OCD

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+## v2.1.0
+-   Add new component for `<pxb-user-menu` (use @pxblue/angular-themes v5.1.0+ to get PX Blue themes for this component).
```
huayunh, line 4: missed a `>`

PR #118, opened 2020-07-24
Feature/bidirectionality

emclaug2:
- Update components to support bidirectionality.
- Add rtl-addon to storybook demo
- Add options to toggle theme and directionality to showcase demo

Each component that needs some conditional styles based on the directionality of the app will have a RTL scss snippet at the top of their sass file that contains support for RTL languages.

#### Screenshots:
image


demos/showcase/src/app/app.component.html
```diff
@@ -1,13 +1,25 @@
-<pxb-drawer-layout [variant]="getVariant()" (backdropClick)="closeDrawer()">
+<pxb-drawer-layout
+    [variant]="getVariant()"
+    (backdropClick)="closeDrawer()"
+    [dir]="isRtl ? 'rtl' : 'ltr'"
+    [class.rtl]="isRtl"
```
joebochill, line 5: Why do we need the dir attribute and the class? Also, shouldn't we be putting this on the `<body>` instead of the DrawerLayout?

demos/showcase/src/app/app.component.ts
```diff
@@ -85,4 +87,20 @@ export class AppComponent {
...
+            body.classList.add('pxb-blue');
+        } else {
+            body.classList.remove('pxb-blue');
+            body.classList.add('pxb-blue-dark');
+        }
+        this.isDarkMode = !this.isDarkMode;
+    }
+
+    toggleDirectionality(): void {
+        this.isRtl = !this.isRtl;
```
joebochill, line 104: set `dir` on body tag?

joebochill:
I also did not check multi-level drawer indentations, are they good?

components/src/core/utility/utility.scss
```diff
@@ -0,0 +1,6 @@
+@mixin invert {
+    transform: scaleX(-1);
+    -moz-transform: scaleX(-1);
+    -webkit-transform: scaleX(-1);
+    -ms-transform: scaleX(-1);
+}
```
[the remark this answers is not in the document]
emclaug2: I removed these extensions and tested out the bidi in firefox and edge, I think we are fine without them.

demos/showcase/src/app/app.component.ts
emclaug2: I'm still keeping isRtl for convenient inline rtl style rules; there's a few of them that were recently added.

demos/showcase/src/app/app.component.html
emclaug2: You're right about the rtl class, it's not required.

Even though I'm not putting the dir attribute on the body, the overlay component is still being updated in the dom to have its `dir="rtl"`.   As I recall, the theme was the item that was causing issues with the ng overlay container, since when we applied the theme on the app root, the overlay wasn't being styled.  

image

Seems like overlay dir functionality is built into angular _somewhere_.

emclaug2:
To keep the scope of this PR down, I've moved the Drawer border style and InfoListItem color issue to a separate feature request found here: 
https://github.com/pxblue/angular-component-library/issues/120

emclaug2:
> I also did not check multi-level drawer indentations, are they good?

They were not good! But now they are.

emclaug2:
image

demos/showcase/src/app/app.component.html
joebochill: I'm just curious why we are putting the dir on the DrawerLayout instead of on the body if our instructions for doing RTL languages say to do it on body.

joebochill: Very strange that the overlay is picking it up even though none of its parents have RTL set...

demos/showcase/src/app/app.component.ts
joebochill: Yeah, I didn't mean to replace it...I was thinking more of an 'in addition to'

joebochill:
Partial dividers still not working in the Showcase:
image

emclaug2:
> Partial dividers still not working in the Showcase:
> image
> 
> Looks like the subtitles also need to be updated to the newer syntax, but that might be outside the scope here (could add it ot that other enhancement you created)

Geez, I tested out the fix in storybook, but the issue is coming from a style precedence issue.  Placing a mat-divider in a mat-card adds some additional styles onto it.  I've fixed the issue by increasing the specificity of our mat-divider rules so that our styles apply. 

image

demos/showcase/src/app/app.component.html
emclaug2: DrawerLayout uses the Direction service (provided by angular) to detect `dir` attributes in its parent DOM tree.  Ive been testing this out for about an hour now; this directive does not detect changes unless you actually use the `dir` directive.

This means if you use javascript to set the dir attribute on the body, or even the app-root, it is ignored by the Direction service and the Drawer doesn't flip to the other side of the screen.

The DrawerLayout component is the only component that actually relies on the DirectionService.  All of our other bidirectionality styles are strictly through css, so they work as long as a parent element in the DOM tree has the `dir` attribute set.

This issue is also present in our storybook app; to test out RTL, I'm using JS to update the dir attribute.  The `within a DrawerLayout story` is broken because the Direction directive `dir` is not used.

Our instructions should be updated for angular to use the `dir` directive instead of setting the `dir` attribute through js.

So this works: 
<div dir='rtl'> .... </div>

But this won't work:
document.body.dir = 'rtl';

joebochill: > So this works:
> `<div dir='rtl'> .... </div>`
> 
> But this won't work:
> `document.body.dir = 'rtl'; `

🤯  whaaat? These are not the same thing?

emclaug2: These are not the same thing, but only because the angular service `Direction` (which DrawerLayout uses for inline styles) explicitly needs the `dir` directive to be used within a template.

joebochill: Ok, so if we're putting it on DrawerLayout, but somebody else has a project that needs RTL, and they're not using DrawerLayout, they would just have to put in on whatever their top-level item is in their app-root?

emclaug2: Yes that's correct, and unfortunately it cannot be placed directly on their app-root either, like you said it has to be the first element within their app-root.

So this won't work: 
```
    <body class="pxb-blue mat-typography mat-app-background">
        <app-root dir="'rtl'"></app-root>
    </body>
```

components/src/core/hero/hero.component.scss
```diff
@@ -1,3 +1,17 @@
+@import '../utility/utility';
+
+/* RTL */
+[dir='rtl'] .pxb-hero {
+    .pxb-hero-primary-wrapper.pxb-hero-svgIcon svg {
+        transform-origin: top right;
```
joebochill, line 6: We're not flipping hero svg icons?

joebochill: image
image

components/src/core/utility/utility.scss
```diff
@@ -0,0 +1,3 @@
+@mixin invert {
+    transform: scaleX(-1);
+}
```
joebochill, line 3: Quick question about this `@invert` mixin. Does this mean we're auto-rotating all icons that a user passes in? What is their recourse if they are using one that shouldn't be flipped? 

On the react-side, we are only auto-rotating the ones that we use internally and leaving it up to the end user to rotate the ones they want to flip.

emclaug2: Since we're doing these rotations through css, if someone doesn't want their icon flipped they could override the transform that is happening, something like this:

```
<pxb-hero label="Battery" value="Full">
      <battery-progress
            style="transform: none"
            percent="90"
             size="36"
            pxb-primary></battery-progress>
</pxb-hero>
```

components/src/core/hero/hero.component.scss
emclaug2: These svg icons will require a little more attention than others; the approach of just applying a scaleX(-1) and adjusting the transform origin is not working for these.

I got it this close:
image

emclaug2: I'm going to capture this requirement as a separate bug since I'm thinking this will take a while to resolve.

emclaug2: #123 Added

emclaug2:
DrawerLayout bidi story is working in a separate branch that uses our own storybook-rtl addon.  It won't be fixed as part of this PR.

emclaug2:
@huayunh  Can you re-review this PR?

huayunh:
@emclaug2 I temporarily lost my ability to spin it locally (that I lose my access to the Eaton MBP), unless you deploy it over Firebase.

emclaug2:
This is latest storybook, with the new rtl addon: http://pxblue-angular-library.web.app/

emclaug2:
This is the latest showcase: https://pxblue-temp.web.app/

huayunh:
Showcase looks good to me!

Storybook: 

1. Typo in the drawer, Calender -> **Calendar**
image

2. Drawer menu icons are not flipped
image

Interestingly, drawer with full config story has half of the icons flipped

image

3. Technically, the home page should not be influenced by i18n flipping... not a big trouble. 
You will get this if you flip the direction in another story then go back to the home page
image

4. The following icons should not be flipped: icons with text (e.g., English letters) and with a particular symbol that crooks a widely recognized symbol when flipped (e.g., earth globe)
image

5. ScoreCard with actions not flipped

image

6. ScoreCard with score badge, card body list is not flipped and misaligned

image

these two points also apply to scorecards with full config story

emclaug2:
The drawer icons look fine to me, which ones are wrong?

image

huayunh:
Something must have slipped in my mind... Ignore the comments about drawers :P

emclaug2:
> Something must have slipped in my mind... Ignore the comments about drawers :P

Everything else makes sense and is being worked on 👍

emclaug2:
#6 is something that will be easier to fix when our custom storybook RTL addon is integrated.  
Turns out we cannot do attribute-based css inside of our storybook custom styles per story.  
We will require an inline fix & an easily accessible `getDirection` function to adjust these scorecard margins.

I think we should also revisit #3 after our own RTL addon is integrated, same reason as above.

emclaug2:
Typo in the Drawer menu items is fixed, scorecard action items are inverted now, and the B hero icon won't be inverted anymore.  Changes are used out here: https://pxblue-angular-library.web.app

PR #125, opened 2020-07-28
Feature/rtl addon

emclaug2:
- Use our own RTL storybook addon, fixes some jankiness of current RTL addon.
- Fix typo in a story
- RTL support for InfoListItem chevron
- Make Welcome page in storybook RTL always
- Adding missing action-row to some scorecard stories, adjust padding to be RTL friendly

Updated storybook: https://pxblue-angular-library.web.app/


jeffvg:
should all the score card chevron(s) be flipped on RTL?

image

emclaug2:
> should all the score card chevron(s) be flipped on RTL?

You're right, I'll put in a fix shortly.

emclaug2:
> > should all the score card chevron(s) be flipped on RTL?
> 
> You're right, I'll put in a fix shortly.

A fix is pushed out now.  Turns out each of the InfoListItems in the ScoreCard stories weren't using the [chevron] prop, they were manually passing in a chevron mat-icon as it's `pxb-right-content`, so it wasn't being inverted when RTL was enabled.   Using the [chevron] prop fixes this issue since that inversion logic is baked into our component.

emclaug2:
Firebase deployment demo up-to-date with latest push.

jeffvg:
For me the user menu with basic usage RTL toggle not flipping icons left or right in menu.
If you toggle the user menu with non text avatar first as RTL then open the basic usage one it will be switched over to RTL. weird...

emclaug2:
> For me the user menu with basic usage RTL toggle not flipping icons left or right in menu.
> If you toggle the user menu with non text avatar first as RTL then open the basic usage one it will be switched over to RTL. weird...

You're right, the UserMenu is giving some strange behavior because its overlay is contradicting the iframe body's dir settings.    This is weird, I'll fix this issue in a bit. 

image

jeffvg:
> > For me the user menu with basic usage RTL toggle not flipping icons left or right in menu.
> > If you toggle the user menu with non text avatar first as RTL then open the basic usage one it will be switched over to RTL. weird...
> 
> You're right, the UserMenu is giving some strange behavior because its overlay is contradicting the iframe body's dir settings. This is weird, I'll fix this issue in a bit.
> 
> image

Seeing this in the toolbar menu example too

huayunh:
Flip icons

image

huayunh:
Everything else looks good to me

emclaug2:
> Flip icons
> 
> image

These icons should already be flipped: 

image

emclaug2:
The DrawerLayout story styles are updated now.

The UserMenu and overlay components will have to be fixed as part of the RTL addon itself.  The addon is not setting the `dir` attribute in the overlay-panel.

I opened an issue in the storybook-addons repo to capture this work.

emclaug2:
The scorecard images are inverted now too and the firebase deployment is updated. 
If you're using the demo site, please open in incognito mode to ensure no caching happens.

PR #127, opened 2020-08-05
Make it so that the drawer text wont show when the drawer is closed

### LINKED ISSUE #126: drawer multiple nav groups example with string not hidden when collapsed
#### Describe the bug
We have 2 title nav items without icons and when drawer is collapsed the first character is displayed. 

To hide the 2 single title nav items on collapse or pass in icon for title to display icon when drawer is collapsed.

1. Go to https://pxblue-components.github.io/angular-dev/?path=/story/components-drawer--with-multiple-nav-groups
2. Click on drawer menu to collapse.

image

emclaug2:
Fixes #126 

- When the drawer is closed, the nav-items with no icons are set to hide.

An alternative solution would be to make this drawer non-closable. 

#### Screenshots:
image

image

### DIFF
--- a/demos/storybook/stories/drawer/with-multiple-nav-groups.stories.ts
+++ b/demos/storybook/stories/drawer/with-multiple-nav-groups.stories.ts
@@ -35,7 +35,7 @@ export const withMultiNavGroups = (): any => ({
               <pxb-spacer *ngIf="spacer"></pxb-spacer> 
               <pxb-drawer-nav-group [title]="groupTitle2" [divider]="true">
                  <pxb-drawer-nav-item *ngFor="let navItem of navItems2"
-                    [title]="navItem.title"
+                    [title]="state.open ? navItem.title : ''"
                     [hidePadding]="true"
                     [selected]="state.selected === navItem.title"
                     (select)="navItem.onSelect(); setActive(navItem.title, state);">
--- a/demos/storybook/stories/drawer/with-multiple-nav-groups.stories.ts
+++ b/demos/storybook/stories/drawer/with-multiple-nav-groups.stories.ts
@@ -6,11 +6,13 @@ export const navItems2 = [
     {
         title: 'Contact',
         itemID: 'group2_item1',
+        icon: 'contact_page',
         onSelect: action('Contact'),
     },
     {
         title: 'Favorites',
         itemID: 'group2_item2',
+        icon: 'favorite',
         onSelect: action('Selected: Favorites'),
     },
 ];
@@ -35,7 +37,7 @@ export const withMultiNavGroups = (): any => ({
               <pxb-spacer *ngIf="spacer"></pxb-spacer> 
               <pxb-drawer-nav-group [title]="groupTitle2" [divider]="true">
                  <pxb-drawer-nav-item *ngFor="let navItem of navItems2"
-                    [title]="state.open ? navItem.title : ''"
+                    [title]="navItem.title"
                     [hidePadding]="true"
                     [selected]="state.selected === navItem.title"
                     (select)="navItem.onSelect(); setActive(navItem.title, state);">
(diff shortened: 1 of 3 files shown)


joebochill:
@huayunh is this the desired behavior? Or do we want to do something like ellipse the text?

emclaug2:
> @huayunh is this the desired behavior? Or do we want to do something like ellipse the text?

I would imagine that we shouldn't be encouraging users to have text-only nav items with a persistent drawer variant.

emclaug2:
Another potential solution here is to just add icons to the second nav group.

joebochill:
Yes, we should encourage them to use icons at the top level with persistent drawer, but we also should define what the behavior will be (clipping, hiding, ellipsing, etc.) if they don't.

huayunh:
(1) Whoever tries to design a product using a persistent drawer without icons has a terrible design taste
(1.1) using a mix of menu items with & without icons on permanent & temporary drawers might be a bad idea too, but not as horrible as (1)
(2) storybook should probably show our best design practice, so set this drawer to permanent or temporary would be a good idea for this story in particular
(3) if they somehow desperately want to use a persistent drawer but with iconless items, what I see in the screenshots above is acceptable; if you want to go fancier, you can shrink the text to caption but font-weight 600 (can be hard to see under 400), and use an ellipsis if still overflowing

image

joebochill:
I'm fine with just hiding it...simplest and easiest to do. Primary design recommendation will be to avoid doing this when possible. Since we haven't introduced the variants yet by this example (and persistent is the default), I'd be fine with just adding icons to the second group in this example. I'm also not opposed to making the variant permanent, it just might be a bit weird for users if they haven't read the readme.

demos/storybook/stories/drawer/with-multiple-nav-groups.stories.ts
```diff
@@ -35,7 +35,7 @@ export const withMultiNavGroups = (): any => ({
               <pxb-spacer *ngIf="spacer"></pxb-spacer> 
               <pxb-drawer-nav-group [title]="groupTitle2" [divider]="true">
                  <pxb-drawer-nav-item *ngFor="let navItem of navItems2"
-                    [title]="navItem.title"
+                    [title]="state.open ? navItem.title : ''"
```
joebochill, line 38: Wait a second...shouldn't we be making this change in the component behavior?

joebochill:
Assuming we'll also want to make this change for React

demos/storybook/stories/drawer/with-multiple-nav-groups.stories.ts
emclaug2: This story needs to updated to add icons to the second nav group.

joebochill: Yes. I think we also need to update the component too. The bug in question is 'labels getting chopped off when collapsed' - so shouldn't we update the component to hide the labels on collapse if there are no icons?

emclaug2:
I'm still unsure what our recommended approach is supposed to be, should we hide the text or shrink the text?

joebochill:
> I'm still unsure what our recommended approach is supposed to be, should we hide the text or shrink the text?

Definitely hide the text

emclaug2:
Drawer no icon

huayunh:
Maybe opacity animation transition instead?

emclaug2:
> Maybe opacity animation transition instead?

And keep the ellipsis?

emclaug2:
> > Maybe opacity animation transition instead?
> 
> And keep the ellipsis?

The ellipsis is something that affects all nav items at the moment; this would be a change for all items.

joebochill:
If it's already doing the ellipsis, I would leave it alone

emclaug2:
> If it's already doing the ellipsis, I would leave it alone

I tried to slow down the Drawer animation, it's using ellipsis currently.  
Ellipis

emclaug2:
d 55

emclaug2:
React has the same bug, made a new issue for it.

Issue #82, opened 2020-06-11
Fix Drawer Jankiness


emclaug2:
#### Describe the bug
When resizing a DrawerLayout component so that it transitions from a persistent variant to a temporary one, the margin-left of the `side nav content` is sometimes incorrect.

#### What is the expected behavior?
Transitions between variants are seemless.

#### What are the steps to reproduce?
1. Open showcase demo.
2. Make screen skinny until variant change is triggered.

#### Screenshots or links to minimum reproduction example
https://github.com/pxblue/angular-component-library/pull/80#issuecomment-642109526

#### What is your environment?

 Framework
- [x] Angular
- [ ] React
- [ ] Ionic
- [ ] React Native

 Browser
- [ ] Internet Explorer
- [ ] Edge
- [ ] Safari
- [x] Chrome
- [ ] Firefox

 Operating System
- [ ] MacOS
- [x] Windows
- [ ] Other

emclaug2:
This bug description is out of date now; in previous bug squashes, this must have been mostly addressed.    There is one pending janky behavior: 

Transitioning from a closed persistent drawer to a closed temporary drawer shows an unnecessary animation. 
![janky](https://user-images.githubusercontent.com/6538289/90631846-6e2c8200-e1f1-11ea-8a07-efd512472ab9.gif)

PR #132, opened 2020-08-06
Update state management logic

emclaug2:
- Refactor State management of the Drawer `open` prop so that our component will not override what the parent component passes in as `@Input`

### DIFF
--- a/components/src/core/drawer/drawer-body/drawer-body.component.ts
+++ b/components/src/core/drawer/drawer-body/drawer-body.component.ts
@@ -7,7 +7,7 @@ import { DrawerService } from '../service/drawer.service';
     changeDetection: ChangeDetectionStrategy.OnPush,
     encapsulation: ViewEncapsulation.None,
     template: `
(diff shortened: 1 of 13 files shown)


emclaug2:
> Looks good, I'll clone it down and play with it tomorrow. Does this by any chance fix the jankiness bug we were seeing when transitioning between variants too? #82

I'll add it to the list of 15 👍

emclaug2:
That story #82 might need updated a bit; I think the Drawer jankiness is partially fixed.

joebochill:
So I'm playing with the showcase. I start it and it loads with the (persistent) drawer open. I resize the window down and it switches to the (temporary) drawer closed. And when I resize it back up, the (persistent) drawer is closed. Shouldn't these all be open since the open prop hasn't changed?

Also seeing an error in the console:
image

emclaug2:
I'm going to make a separate issue for the console error you're seeing above because it looks like it's stemming from the toolbar component.

emclaug2:
This issue is fixed now - the showcase demo was manually setting the `open` state to false when transitioning from a `temporary` to `persistent` variant.

PR #136, opened 2020-08-08
Add icon align prop

huayunh:
Fixes PXBLUE-1472

- Add `iconAlign` attr

#### Screenshots:

image


daileytj:
Not sure how big of a deal this is, but when the page is RTL and iconAlign="left", the icon is aligned to the right. This is also the case for iconAlign="right"... the icon is then left aligned for RTL.

CHANGELOG.md
```diff
@@ -1,29 +1,34 @@
 # Change Log
 
+## v2.1.1
```
joebochill, line 3: This can go in 2.1.0 (not released yet)

joebochill:
> Not sure how big of a deal this is, but when the page is RTL and iconAlign="left", the icon is aligned to the right. This is also the case for iconAlign="right"... the icon is then left aligned for RTL.

We discussed this on the React PR and agreed that it's fine. We already have things like 'leftContent' and 'rightContent' which will be backwards too. Given that the primary use case is LTR, we will orient these props based on that.

PR #142, opened 2020-08-11
Invert svgIcons and use barley icon

### LINKED ISSUE #123: Hero SVG Icons not Flipped for RTL languages
#### Describe the bug
The hero component is not flipping svg when testing RTL languages.

When RTL is enabled, all icons passed into the HeroComponent should be inverted.

1. Run the showcase, replace the GradeA icon in the HeroBanner with a Barley icon and observe.

LTR: 
image

RTL:
image

emclaug2:
Fixes #123 

- Update icon types story to use the `barley` icon instead of `grade_a`.
- Apply invert logic to svgIcons too

#### Screenshots:
barley

### DIFF
--- a/components/src/core/hero/hero.component.scss
+++ b/components/src/core/hero/hero.component.scss
@@ -2,11 +2,8 @@
 
 /* RTL */
 [dir='rtl'] .pxb-hero {
-    .pxb-hero-primary-wrapper.pxb-hero-svgIcon svg {
-        transform-origin: top right;
-    }
     .pxb-hero-primary-wrapper {
-        &:not(.pxb-hero-svgIcon) > * {
+        > * {
             @include invert();
         }
     }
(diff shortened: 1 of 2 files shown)


jeffvg:
is it ok for the top and bottom of image to bleed out?

image

emclaug2:
This is tricky because the current scaling for these svg icons targets icons that don't span from edge-to-edge of the svg container.   We can always make the icons a little smaller to account for large icons, but then icons that previously looked fine (grades) will now be smaller.

joebochill:
We definitely do not want the icons sticking out past the background...do we have an issue logged for that?

Also, what happens now if we use a hero with a text/span as the banner graphic?

jeffvg:
> We definitely do not want the icons sticking out past the background...do we have an issue logged for that?
> 
> Also, what happens now if we use a hero with a text/span as the banner graphic?

logging separate issue shortly....

PR #144, opened 2020-08-20
Add rail

emclaug2:
- Adds `rail` variant.  (name pending)
- Check it out [here](https://pxblue-angular-library.web.app/?path=/story/components-drawer--within-a-drawer-layout&knob-width=350&knob-variant=rail).

#### Screenshots:
image


emclaug2:
How do the styles look @joebochill  @huayunh ?

emclaug2:
Any thoughts on the animation from open to closed?  The text is set to just appear right now; it's a very stiff transition.

emclaug2:
Also following the DSM for the nav-rail, the height of each nav-item is 56px.  Normally, each nav-item is 52px.   

Nav-items are supposed to be 52px high, right?

huayunh:
I think there should be no margins between text and their icons. Let me check those out with Kyle...

emclaug2:
> I think there should be no margins between text and their icons. Let me check those out with Kyle...

Can you also ask about the line-height between the words "Identity" and "Management".    I think the line-height should match the font-size.

joebochill:
High-level initial thoughts:
- Text is too small. Material recommends 14dp minimum and cautions against wrapping labels.
    - This has been my biggest hesitation with us building a Rail variant, especially if it's going to be used in multilingual apps
- My expectation was that the rail variant does not open and close, it's just supposed to be a rail. That's at least how Material describes it.
- Refer to the Material specs: https://material.io/components/navigation-rail

huayunh:
Echo what joe said above -- material wants their rail to be 72dp wide. I am checking with Kyle to see why they made that 56dp in DSM.

joebochill:
> Echo what joe said above -- material wants their rail to be 72dp wide. I am checking with Kyle to see why they made that 56dp in DSM.

In the spec section, they do a 56x56 condensed version with icons only (no labels).

huayunh:
yup... but the DSM version comes with both icons and labels

emclaug2:
I'm thinking the expectation for the nav rail is that there should be not header, since we can't open/close the drawer.

image

emclaug2:
The firebase demo is updated with latest changes. 

NOTE: Storybook is having some weird change-detection errors when switching between rail and any other variant.   Switch variants, then try opening/closing drawer or selecting nav items.

emclaug2:
Are we adding support for dense nav rail?

image

emclaug2:
Also, theme needs to be updated to change `selected` text color when using the `rail` variant.

joebochill:
I am okay with skipping the dense version for the first iteration of this. I think continuing to support the header would be nice since this can be used as a product branding area by users.

emclaug2:
This is ready for review now.  Here's what the latest version looks like. 

The header area can be used for branding or can be removed entirely.   If there's a small eaton logo that would look good in that space I can swap out the topology image.  

image

components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
```diff
@@ -80,6 +80,22 @@ pxb-drawer-nav-item {
             transform: rotate(180deg);
         }
     }
+    .pxb-drawer-nav-item-rail {
+        display: flex;
+        flex-direction: column;
+        height: 72px;
+        width: 71px;
```
joebochill, line 87: 72?

components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.ts
```diff
@@ -79,9 +81,21 @@ export type ActiveItemBackgroundShape = 'round' | 'square';
                     >
                 </div>
             </pxb-info-list-item>
+            <ng-container *ngIf="isRail()">
+                <div
+                    class="pxb-drawer-nav-item-rail square"
```
joebochill, line 86: Do we want to make 'square' a pxblue-style class name?

emclaug2: Some of our classes are more geared for conditional styling based around internal state.  Originally I was thinking that end users wouldn't be concerned with _every_ class we manage, but it's hard for us to judge which ones to expose.

I think we can make our lives easier by just applying to the `pxb-[component-name]` prefix to every class.  I would make this a separate github issue though because there are a few of these scattered throughout our comp lib.

components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
emclaug2: Yeah this should be 72, ran into some issues with a 1px border causing some overflow.  I'll play around with this more.

emclaug2:
> Looking good. I'm impressed that we're able to get this done in so few code changes 😮 .
> 
> * The active selected colors look off (in showcase)
> * What is the plan for longer labels? I know we will discourage people from using long labels, but what if they do? Material says not to use ellipses in the rail...hyphenate words and use line break instead of overflowing out of the container? should we enforce a character limit?
> * What are people's thoughts about using our Monospace font for these labels? or Overline?
> * Should the labels be bold?
> * Do we want to do anything in the documentation to talk about the restrictions of the rail variant (limited character labels, no nesting of nav items, various properties have no effect like width, open, etc.)?
> * Should the header center it's content in the 72px width for rail?
> 
> image

The theme needs to be updated; it determines what colors are applied when an item is selected.
I think most of these questions will be answered when the rail is added to DSM. 

I'm thinking if an icon or header-content is provided in the DrawerHeader then it should be centered.  I will update the default styles to do this.   All other style questions I think should wait for @huayunh or @kyle's input.

emclaug2:
I've updated the nav rail example to have a placeholder logo image in the header.

Also I've updated the header styles to automatically center a 24x24px icon or custom `pxb-title-content`.

demos/storybook/stories/drawer/within-drawer-layout.stories.ts
```diff
@@ -3,32 +3,46 @@ import * as Colors from '@pxblue/colors';
...
 const items = [...nestedNavGroup];
 
 export const withinDrawerLayout = (): any => ({
+    styles: [
+        `
+        :host { 
+            height: 100%; 
+            width: 100%;
+        }
+        ::ng-deep .pxb-drawer-header-background {
```
joebochill, line 18: Why do we need ng-deep here? I thought our component encapsulation rules let us avoid this.

emclaug2: Storybook style overrides require an ng-deep because each story has the default angular ViewEncapsulation applied to it which would scope each stories' styles to only affect local elements.  

We won't have this problem in the showcase because our demo project has ViewEncapuslation.None.

PR #148, opened 2020-08-28
Feature/round drawer ripple

emclaug2:
- This makes the ripple effect in angular drawer contained to the selected outline.

The fix was to stop using css:after content to draw the highlighted area.  Now the highlight area is a separate div, `.pxb-drawer-nav-item-highlight`, which is positioned with `absolute` behind the content of each nav item.

#### Screenshots:
Circular Ripple:
circle ripple

Square Ripple: 
square ripple


components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
```diff
@@ -55,23 +52,24 @@ pxb-drawer-nav-item {
...
         opacity: 0.9;
-    }
-    .pxb-info-list-item-active.round:after {
-        width: calc(100% - 8px);
-        border-radius: 0 24px 24px 0;
-    }
-    .pxb-info-list-item-active.square:after {
-        border-radius: 0;
-        width: 100%;
+        min-width: 120px;
```
joebochill, line 63: Can we drop the min-width here? If user sets width of the drawer to less than this, the round part get clipped.

emclaug2: The only reason why this is here to band-aid the border radius that appears when selecting/opening an item in the persistent variant.

This is the issue that I was seeing:  When opening the drawer, the rounded border of the selected item arrives too early. 
circle


This is what a 120px drawer looks like: 
image

This fix just tries to mask when the rounded corners show up; I don't think it's a very elegant fix.  Any idea on how to tweak it to make it work with any width?

emclaug2: Another idea I had would be to wait until the drawer is opened (animation over) before applying the `round` class onto the highlight area.

joebochill: You might be able to animate/transition the border radius too. That might help smooth it out.

jeffvg:
I see in the screen shots above you have active item background color set, do we have a knob for that in full config example

joebochill:
> I see in the screen shots above you have active item background color set, do we have a knob for that in full config example

The activeBackground color comes from the theme...it's not something we would encourage user's to customize, but they could do it through classes if they really wanted to.

jeffvg:
> > I see in the screen shots above you have active item background color set, do we have a knob for that in full config example
> 
> The activeBackground color comes from the theme...it's not something we would encourage user's to customize, but they could do it through classes if they really wanted to.

ok wasn't sure, we do expose this in the React example via knob

joebochill:
> > > I see in the screen shots above you have active item background color set, do we have a knob for that in full config example
> > 
> > 
> > The activeBackground color comes from the theme...it's not something we would encourage user's to customize, but they could do it through classes if they really wanted to.
> 
> ok wasn't sure, we do expose this in the React example via knob

Oh right, I forgot that we do expose that as a prop for react. I think in the React case, we're showing everything that's available through props. On Angular, we're showing everything customizable through Inputs. activeItemBackgroundShape is an Input for angular but backgroundcolor is not (it's handled through classes).

joebochill:
Have we tried anything related to animating that highlight color so we don't have to specify the min-width?

emclaug2:
> Have we tried anything related to animating that highlight color so we don't have to specify the min-width?

I played around with the animation to help reduce the smoothing but even that's a bit of a challenge since it only reduces the amount of jankiness, but then we also only want to apply the animation when opening the drawer.  If the drawer is already opened, the animation that happens when selecting items is out of place. 

I'm hoping to have to some time later today to finish up this PR; I don't think the min-width is a terrible solution the more I play around with this problem.   This is assuming nav items titles consist of at least 2 letters.

emclaug2:
> Have we tried anything related to animating that highlight color so we don't have to specify the min-width?

I just removed the min-width & applied a transition instead.  Looks better.

joebochill (approved):
Looks better...do we want to match the same transition easing that the drawer is using (I think right now it's linear):
```width 300ms cubic-bezier(0.4, 0, 0.2, 1) 0ms```

Issue #112, opened 2020-07-14
Components should set display rule for first rendered elements


joebochill:
#### Describe the bug
The first element in our component DOM trees do not always set a display type. This can cause unexpected behavior when trying to style certain components. This is a known issue in the InfoListItem, but we should verify the other components as well.

#### What is the expected behavior?
First element in component trees should explicitly set the desired display type.

#### Anything else to add?
When the browser doesn't recognize a HTML tag (such as for custom components), it renders them with display:inline by default (with height/width of 0 and certain CSS rules having no effect). When the first element in the DOM of a component is also an unrecognized browser tag (such as `<mat-list-item>`), that seems to be when the issues arise - we just need to make sure these root elements have an explicit display type.
Do this with #111

joebochill:
This was related to the drag and drop example where the dragged item styling was getting broken because of the inline display.

This one will not be a breaking change, but #111 is.

Issue #111, opened 2020-07-14
Follow the Angular Material class naming paradigm


joebochill:
#### Describe the desired behavior
Host elements for our components should have a class attached to them. For example, observe the `<mat-list-item>` component from angular material has a class `.mat-list-item`. The first rendered element has a class of `.mat-list-item-content`

#### Describe the current behavior
We are currently applying the `.pxb-*` class to the first element (not the host).

#### Is this request related to a current issue?
This was found as part of the drag/drop issue we uncovered with the sortable list design pattern example. It's also related to #112.

#### Additional Context
We can set classes on the host element using the following syntax in the Component decorator:
```
host: {
    'class': 'pxb-info-list-item-test',
},
```

joebochill:
This is definitely going to be a breaking change. We will need to have some documentation (in the changelog?) about how to migrate if you are using the old class names in your existing application.

PR #149, opened 2020-09-01
Fix banner display settings, add border to banner story

emclaug2:
- Add width knob to HeroBanner story.
- Heroes within hero-banner all set to grow.
- Add border around HeroBanner story.

#### Screenshots:
Screen Shot 2020-09-01 at 2 04 53 PM

### DIFF
--- a/components/src/core/hero/hero-banner.component.ts
+++ b/components/src/core/hero/hero-banner.component.ts
@@ -11,6 +11,9 @@ import { ChangeDetectionStrategy, Component, Input, ViewEncapsulation } from '@a
         <mat-divider class="pxb-hero-banner-divider" *ngIf="divider"></mat-divider>
     `,
     styleUrls: ['./hero-banner.component.scss'],
+    host: {
+        class: 'pxb-hero-banner-root', // TODO: Follow ng mat naming conventions
+    },
 })
 export class HeroBannerComponent {
     @Input() divider = false;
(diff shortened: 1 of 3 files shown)


components/src/core/hero/hero-banner.component.ts
```diff
@@ -11,6 +11,9 @@ import { ChangeDetectionStrategy, Component, Input, ViewEncapsulation } from '@a
         <mat-divider class="pxb-hero-banner-divider" *ngIf="divider"></mat-divider>
     `,
     styleUrls: ['./hero-banner.component.scss'],
+    host: {
+        class: 'pxb-hero-banner-root', // TODO: Follow ng mat naming conventions
```
joebochill, line 15: Can we do the TODO now so we don't have a breaking class change in the future? Is the mat convention to make the host `pxb-hero-banner`?

emclaug2: Yes this TODO is a reference to #111 & #112.   

The TODO can actually be removed since the work is already captured in those issues.

joebochill: I'm suggesting we use the desired class name now instead of setting it to `pxb-hero-banner-root` now (breaking change 1) and then changing it to `pxb-hero-banner` later (breaking change 2).

Unless we're planning to use `-root` as our convention for all of the components, in which case it's fine.

joebochill: I guess either way it's a breaking change though...we really just need to do #111 and #112 sooner than later.

PR #156, opened 2020-09-09
Feature/112 host class

### LINKED ISSUE #111: Follow the Angular Material class naming paradigm

### LINKED ISSUE #112: Components should set display rule for first rendered elements

emclaug2:
Fixes #112, Fixes #111 

- Top-level components have a class applied to the host element
-  Docs updated

UserMenu has no root-class, so it has no `pxb-user-menu-content` class.

These components are internal to the Drawer; I didn't apply a `display: flex` to them since they work fine already and shouldn't be used outside of the Drawer. 
- DrawerHeader
- DrawerSubheader
- DrawerBody
- DrawerNavItem
- DrawerNavGroup
- DrawerFooter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## v2.2.0 (to be published)
+## v3.0.0 (to be published)
 
 -   Adds a new `rail` variant to the `<pxb-drawer-layout>`.
 -   Removed right-to-left icon inverting logic per [Material Design's bidirectionality guideline](https://material.io/design/usability/bidirectionality.html#mirroring-layout).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 -   Adds a new `rail` variant to the `<pxb-drawer-layout>`.
 -   Removed right-to-left icon inverting logic per [Material Design's bidirectionality guideline](https://material.io/design/usability/bidirectionality.html#mirroring-layout).
+-   Adds a host class to each PX Blue component tag
 
 ## v2.1.0
 
--- a/components/src/core/channel-value/channel-value.component.scss
+++ b/components/src/core/channel-value/channel-value.component.scss
@@ -1,10 +1,14 @@
 /* RTL */
-[dir='rtl'] .pxb-channel-value .pxb-channel-value-icon-wrapper {
+[dir='rtl'] .pxb-channel-value-content .pxb-channel-value-icon-wrapper {
     margin-right: 0;
     margin-left: 4px;
 }
 
 .pxb-channel-value {
+    display: inline-flex;
+}
+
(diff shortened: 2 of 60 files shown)


emclaug2:
Themes are backwards-compatible; no changes needed.

joebochill:
> Themes are backwards-compatible; no changes needed.

Really? That's nice.

emclaug2:
> > Themes are backwards-compatible; no changes needed.
> 
> Really? That's nice.

Yep; our themes are just using css with the parent class of each component already assumed to be `pxb-[component-name]`, so this refactor works pretty nicely.  

The only issue that I've noticed is that when I changed the host element of the DrawerLayout to be flex, the body of the showcase app needed to specify a height in order to work.   This is a bit annoying.

huayunh:
Changelog?

huayunh:
code looks all good

joebochill (commented):
Looks good. 

This is still a breaking change though even though the themes are backward compatible right? Should we update the changelog to be 3.0.0?

Also, we probably want to set the root display for the ILI to block instead of flex, or the scorecard footers look goofy:
image

emclaug2:
> Looks good.
> 
> This is still a breaking change though even though the themes are backward compatible right? Should we update the changelog to be 3.0.0?
> 
> Also, we probably want to set the root display for the ILI to block instead of flex, or the scorecard footers look goofy:
> image

This stuck out to me too but turns out the showcase has some styles that are overriding the default behavior of the scorecard's InfolistItem.    I just opened a PR to update the showcase here: https://github.com/pxblue/angular-showcase-demo/pull/2

joebochill:
> This is still a breaking change though even though the themes are backward compatible right? Should we update the changelog to be 3.0.0?

joebochill (approved):

PR #157, opened 2020-09-10
Add height and width specs to the pxb-drawer-layout-content wrapper

emclaug2:
- Add styles to specify height/width

#### Screenshots:
This fix removes the gray area as seen here:
image


emclaug2:
Let's get the #156 merged in before addressing style issues.

PR #159, opened 2020-09-15
Feature/navrail match dsm

### LINKED ISSUE #160: DrawerLayout pxb-content not 100% height
#### Describe the bug
The content area of the DrawerLayout should expand to the bottom of the page.

A user should not have to specify that their content expands 100vh.

Go to the design patterns landing page.

image

#### Environment

Windows/Chrome
		
#### Anything else to add?
No

emclaug2:
Fixes #160 

-  Update rail variant to match DSM specs
-  Add condensed option to rail variant, which hides title and adds a tooltip.
-  Fix bug where not all of nav item was being rippled onclick.

#### Screenshots:
d


emclaug2:
Nav Items do not grow in size based on the size of the content inside.  Rail items are strictly 72 x 72px. 
image

emclaug2:
`hyphens: auto` does not work on chrome on windows machines.   Since most target users at Eaton are on windows I'm omitting the hypen word-break.

https://caniuse.com/?search=hyphens

emclaug2:
TODO: Changelog & Update API docs

huayunh:
> Nav Items do not grow in size based on the size of the content inside. Rail items are strictly 72 x 72px.

So we cannot use some vertical padding to make it 72px * 72px min-height and expand vertically when the text wraps?

emclaug2:
> > Nav Items do not grow in size based on the size of the content inside. Rail items are strictly 72 x 72px.
> 
> So we cannot use some vertical padding to make it 72px * 72px min-height and expand vertically when the text wraps?

I think I misinterpreted what the story was asking to me do; do you want the nav items to dynamically expand their height if the content does not fit?  Right now it's a static 72x72px because I thought it should not grow.

components/src/core/drawer/drawer-layout/drawer-layout.component.scss
```diff
@@ -38,6 +38,7 @@ $transition: 300ms cubic-bezier(0.4, 0, 0.2, 1);
 
     .pxb-drawer-layout-sidenav,
     .pxb-drawer-layout-nav-content {
+        height: inherit;
```
emclaug2, line 41: This fixes #160

huayunh:
> do you want the nav items to dynamically expand their height if the content does not fit?

Yes, that would be ideal

emclaug2:
I've updated the height of the nav-items to use min-heights.  This is what the items look like when there's a large amount of text: 

image

jeffvg (approved):
looks good

PR #161, opened 2020-09-17
Add font-size to info-list-item subtitle and info wrappers

daileytj:
Fixes PXBLUE-1537.

- Add font-size to info-list-item subtitle and info wrappers

Notes: This update is being made so that the angular-deign-pattern repo has the appropriate font-sizes for subtitle and info content.

Follow up tasks:
- publish new @pxblue/angular-components beta package
- update angular-design-pattern repo with latest @pxblue/angular-components beta package

### DIFF
--- a/components/src/core/info-list-item/info-list-item.component.scss
+++ b/components/src/core/info-list-item/info-list-item.component.scss
@@ -95,13 +95,16 @@ $denseHeight: 52px;
         margin-bottom: 0;
     }
 
-    .pxb-info-list-item-subtitle-wrapper {
+    .pxb-info-list-item-subtitle-wrapper,
+    .pxb-info-list-item-info-wrapper {
         font-weight: 400;
         line-height: 1.3;
+        font-size: 14px;
     }
 
     .pxb-info-list-item-title-wrapper,
-    .pxb-info-list-item-subtitle-wrapper {
+    .pxb-info-list-item-subtitle-wrapper,
+    .pxb-info-list-item-info-wrapper {
         * {
             white-space: nowrap;
             overflow: hidden;


emclaug2:
Is there a reason why we're specifying 14px?  The subtitle and info is already size 14px because the `mat-body-2` typography is applied on the subtitle & info wrapper.

jeffvg (approved):

daileytj:
> Is there a reason why we're specifying 14px? The subtitle and info is already size 14px because the `mat-body-2` typography is applied on the subtitle & info wrapper.

This was the only way I could consistently get things working in the design-pattern repo. I'm all ears for a better solution. The mat-body-2 should indeed be setting the font-size but for some reason it's not working in the design-pattern project. I can keep digging if we don't like this approach.

emclaug2:
> > Is there a reason why we're specifying 14px? The subtitle and info is already size 14px because the `mat-body-2` typography is applied on the subtitle & info wrapper.
> 
> This was the only way I could consistently get things working in the design-pattern repo. I'm all ears for a better solution. The mat-body-2 should indeed be setting the font-size but for some reason it's not working in the design-pattern project. I can keep digging if we don't like this approach.

Is this something that is being observed on stackblitz or locally or both?  I'm really curious to see what's going on here before merging this PR in.

emclaug2:
This looks good to me; this issue is caused by angular material v9.  In 9, they introduced more specific css rules for mat-list-items that are overriding our `mat-body-2` font-size.  This is why we see this issue in the design pattern repo but not in storybook.

emclaug2 (approved):

PR #164, opened 2020-09-22
Add display flex to host class

emclaug2:
- Add `display: flex` to host class on UserMenu

If the host class (or parent container) does not have display: flex specified, the default width is 100% of the container.  It'll look something like this:

image

Adding display: flex to the host class makes the UserMenu look normal again:
image


emclaug2:
After this gets merged in, I can publish a new beta package for Nikita.

daileytj (approved):

PR #165, opened 2020-09-22
Feature/nav rail enhancements

emclaug2:
- Add `condensed` rail documentation
- Show what a word-break looks like on nav rail.
- Show what a dash delimited break looks like on nav rail.
- Update DrawerLayout example rail header to match rail example.

#### Screenshots:
image


jeffvg:
didn't notice this before but the icons are not switching over when RTL

huayunh:
It was my bad before. For when to mirror what icon, see https://material.io/design/usability/bidirectionality.html#mirroring-layout

huayunh:
I've been noticing this strange ripple behavior that sometimes the ripple does not cover the entire box before it vanishes. The exact triggering condition is unknown.

ezgif-1-d88da4f82bc3

huayunh:
What was the rationale behind making the menu item padding `14px 8px 16px` instead of `15px 8px`?

docs/Drawer.md
```diff
@@ -23,9 +23,13 @@ Parent element (`<pxb-drawer>`) attributes:
...
 
-| @Input | Description          | Type      | Required | Default |
-| ------ | -------------------- | --------- | -------- | ------- |
-| open   | State for the drawer | `boolean` | yes      |         |
+| @Input    | Description                            | Type      | Required | Default |
+| --------- | -------------------------------------- | --------- | -------- | ------- |
+| open      | State for the drawer                   | `boolean` | yes      |         |
+| condensed | Skinny view for `rail` variant         | `boolean` | no       | false   |
+
+> ** The `condensed` attribute won't have any effect on the `<pxb-drawer>` unless the `rail` variant is set on the `<pxb-drawer-layout>` component.  Each item in a navigation rail will be sized 72 x 72px.  When using a `condensed` rail, each item will be sized 52 x 52px.
```
huayunh, line 31: 52x52 or 56x56?

emclaug2: Good catch this should be 56 x 56

emclaug2:
> What was the rationale behind making the menu item padding `14px 8px 16px` instead of `15px 8px`?
> 
> Edit: the design says the horizontal padding should be 4px

I was following material design's specs for their nav rail found here: https://material.io/components/navigation-rail#specs

emclaug2:
> I've been noticing this strange ripple behavior that sometimes the ripple does not cover the entire box before it vanishes. The exact triggering condition is unknown.
> 
> ezgif-1-d88da4f82bc3

I've noticed this too but couldn't figure out what was causing this strange half-fill effect.  Seeing that it's seemingly random, I was just going to ignore it.   Maybe we could put an issue for it?  I'm pretty sure the issue has something to do with our usage of the ripple instead of the mat-ripple itself.

huayunh:
I see. It looks like the aligning is different here. They are aligning up to the text's baseline, rather than where the text actually ends.

image

huayunh:
(which they often do in their docs... their docs can be full of mines)

emclaug2:
I changed the padding to be 15px 8px; it looks fine to me.

image

huayunh:
Looking right.

huayunh:
wait can you make selected item font weight 600? Or does this go with other drawer updates?

emclaug2:
> wait can you make selected item font weight 600? Or does this go with other default drawer style updates?

I would save this for the general drawer styles story, since it's outside scope of the rail.

huayunh:
Sounds good! Looks nice to me otherwise

joebochill:
> I've been noticing this strange ripple behavior that sometimes the ripple does not cover the entire box before it vanishes. The exact triggering condition is unknown.

It happens when you click on an item when something else is focused. Once your first click focuses the new item, clicking it again uses the full ripple, but if you click another one, that one will have the partial ripple. It might be by design, idk.

demos/storybook/stories/drawer/with-nav-rail.stories.ts
```diff
@@ -5,19 +5,19 @@ const headerImg = require('../../assets/eaton-condensed.png');
 
 export const navItems: DrawerNavItem[] = [
```
joebochill, line 6: I'm not sure how you guys feel about this, but should we update these items/icons in the Drawer to be things that you might actually see in a drawer? E.g.:
- Overview
- Alarms
- Settings
- Device Settings

joebochill (commented):
On the Drawer Layout story, the active state isn't set until you move your mouse off of the selected item...looks like a small change detection issue.

demos/storybook/stories/drawer/with-nav-rail.stories.ts
emclaug2: How's this?

image

This is using the 16px padding as well.

emclaug2: @huayunh @joebochill Is there a EATON svg for the DrawerHeader?  The image I'm using looks a bit blurry.

huayunh: Have to publish @pxblue/icons first for that eaton thing

The eaton logo is weird to be on the top, bc usually you'd expect those in the footer. Maybe use a pxblue logo instead and make it look more like a header (like using a blue background)?

huayunh: Kyle might have made some recommendations on what to use for those menu items in his storybook audit. I couldn't locate the miroboard though....

components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
```diff
@@ -98,7 +98,7 @@ pxb-drawer-nav-item {
         width: 72px;
         justify-content: center;
         align-items: center;
-        padding: 14px 8px 16px 8px;
+        padding: 15px 8px;
```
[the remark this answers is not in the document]
huayunh: As long as it does not bump up the 72px height

demos/storybook/stories/drawer/with-nav-rail.stories.ts
emclaug2: I'm really glad you mentioned this @huayunh.  I moved the header content to the footer and found a bug.  The footer always thinks the drawer is closed and sets the visibility of the rail footer to hidden.  I'm pretty sure this issue applies to the subheader as well.  

Since some team(s?) have expressed interest in having conditional logic in the footer based on whether the drawer is opened/closed, does it make sense to remove the visibility logic from our drawer?

A user can pretty easily hide content themselves by just putting `*ngIf="open"` on the `<pxb-footer>`.  Thoughts @joebochill  @huayunh ??

emclaug2: My 2c I'm thinking to remove the conditional visibility from our drawer.   I can do this refactor as part of this PR if this makes sense.

components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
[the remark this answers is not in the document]
emclaug2: This change won't affect the 72x72 or 56x56 condensed height.

emclaug2:
I've moved the nav rail header to the footer for each example. 

image


image

demos/storybook/stories/drawer/with-nav-rail.stories.ts
joebochill: I would not take the logic out until we have implemented a way for them to opt-in/opt-out. Otherwise it will break for people who are currently relying on it and we don't have a way for them to re-enable it (without doing it themselves).

joebochill (approved):

demos/storybook/stories/drawer/with-nav-rail.stories.ts
emclaug2: This could be a possible item to consider in the future; it'd have to be for all the Drawer implementations as well.  We can revisit this if teams are finding this feature inconvenient.

huayunh:
Is there a directive to turn off the nav rail tooltip?

emclaug2:
> Is there a directive to turn off the nav rail tooltip?

No, whenever the `condensed` prop is active, the tooltip is currently not optional.  If this is something we want users to have toggle-able we can include it in the Drawer refactor story.

PR #167, opened 2020-09-28
Merge latest changes into master

joebochill:
- Merge the latest updates into master

> published as @pxblue/angular-components@3.0.0


daileytj (approved):

daileytj:
Do we have a good way of communicating the root class ```...-content``` updates since these would cause breaking changes for anyone using the old classes for custom styles?

joebochill:
> Do we have a good way of communicating the root class `...-content` updates since these would cause breaking changes for anyone using the old classes for custom styles?

@evan suggested that the old classes would still work (I think). I think they'd be styling a different element, but the effect is the same?

Is it mentioned in the changelog?

emclaug2:
There's a mention of the new host class in the changelog @joebochill  You're right that anyone using the old top-level class names to add styles will now be styling a new element, but this shouldn't matter too much.  Our older themes will style the adjusted components just fine.

emclaug2 (approved):

huayunh (commented):
Can we change the list item tag to match what's in the DSM? Or is that for the next release?

The latest design is font size 10px, font weight 700, letter-spacing 1px, line-height 16px

joebochill:
> Can we change the list item tag to match what's in the DSM? Or is that for the next release?
> 
> The latest design is font size 10px, font weight 700, letter-spacing 1px, line-height 16px

Next release

PR #169, opened 2020-09-29
Add compat table and update rtl-addon version

emclaug2:
- Adds compatibility table to README
- Update rtl-addon version

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,10 @@ yarn link:components
 
 The following table outlines which versions of Angular and PX Blue resources will work together.
 
-| @angular         |  @pxblue/angular-components | @pxblue/angular-themes |
-| ---------------- | --------------------------- | ---------------------- |
-| `^8.0.0`         | `^3.0.0`                    | `^5.1.2`               |
-| `^9.0.0`         | `^3.0.0`                    | `^5.1.2`               |
+|  @pxblue/angular-components   |  @pxblue/angular-themes     | @angular + @angular/material |
+| ----------------------------- | --------------------------- | ---------------------------- |
+| `^2.0.0`                      | `^5.0.0`                    | `^8.0.0`                     |
+| `^3.0.0`                      | `^5.1.0`                    | `8.x - 9.x`                  |
 
 ## Running the demo projects
 
--- a/README.md
+++ b/README.md
@@ -40,6 +40,15 @@ There is also a demo project (/component-demo) in this repository that shows the
 yarn link:components
 ```
 
+## Compatibility
+
+The following table outlines which versions of Angular and PX Blue resources will work together.
+
+| @angular         |  @pxblue/angular-components | @pxblue/angular-themes |
+| ---------------- | --------------------------- | ---------------------- |
+| `^8.0.0`         | `^3.0.0`                    | `^5.1.2`               |
+| `^9.0.0`         | `^3.0.0`                    | `^5.1.2`               |
+
 ## Running the demo projects
 
 This repository comes with two demo projects found within the `/demos` folder.
--- a/demos/storybook/package.json
+++ b/demos/storybook/package.json
@@ -30,7 +30,7 @@
         "@pxblue/icons": "^1.0.25",
         "@pxblue/icons-svg": "^1.0.18",
         "@pxblue/ng-progress-icons": "^1.1.4",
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -40,6 +40,15 @@ There is also a demo project (/component-demo) in this repository that shows the
...
 
+## Compatibility
+
+The following table outlines which versions of Angular and PX Blue resources will work together.
+
+| @angular         |  @pxblue/angular-components | @pxblue/angular-themes |
+| ---------------- | --------------------------- | ---------------------- |
+| `^8.0.0`         | `^3.0.0`                    | `^5.1.2`               |
+| `^9.0.0`         | `^3.0.0`                    | `^5.1.2`               |
+
```
joebochill, line 51: Don't we also know that the last version of the themes/components will work with Angular 9 since we just published these today?

Shouldn't these be 

```suggestion
| @angular         |  @pxblue/angular-components | @pxblue/angular-themes |
| ---------------- | --------------------------- | ---------------------- |
| `8 - 9`         | `^2.0.0`                    | `^4.0.0`               |
```

joebochill: This is how I'm used to seeing these:
image

emclaug2: > Don't we also know that the last version of the themes/components will work with Angular 9 since we just published these today?

The issue I'm seeing is when when people start trying to use @pxblue/angular-components@3.0.0 with @pxblue/angular-themes@4.0.0

Themes@4.0.0 won't have styles for new components introduced in components@3.0.0.  I think limiting the range reduces confusion.  

New themes are backwards compatible with older component libraries, so we could do this:


| @angular         |  @pxblue/angular-components | @pxblue/angular-themes |
| ---------------- | ----------------------------------- | ---------------------------- |
| `8 - 9`               | `>=2.0.0`                                     | `>=5.0.0`                            |

joebochill: To me that's a little confusing. Version 2 of the component library does not require version 5 of the theme does it?

Would it make more sense to base this off of the 'package you are viewing', e.g.:
| @pxblue/angular-components  |  @pxblue/angular-themes | @angular + @angular/material |
| ------------------------------- | ------------------------- | ------------------------------ |
| `^1.0.0`                                        | `>=4.0.0`                          | 8                                                |
| `^2.0.0`                                        | `>=5.0.0`                          | 9                                                |
| `^3.0.0`                                        | `>=5.1.0`                          | 9                                                |

emclaug2: I like this approach, but I still think people will read the table from right-most to left-most.   The angular version will be what drives the version of @pxblue they can use, not the otherway around.  

I'm happy with this format if you're happy  👍

emclaug2: > To me that's a little confusing. Version 2 of the component library does not require version 5 of the theme does it?

I don't think we should advertise using older versions since we dont have an easy way to confirm it's working.   I can't remember what version 4.0.0 of the themes looks like.

emclaug2: I like the table above, but ^1.0.0 doesn't use angular 8.  I think it used 6, but I can't remember clearly.  

I would avoid even putting 1.0.0 on this chart since I've got no idea what version of themes it used.

joebochill: Yeah I was just slapping some numbers in there as an example to show a progression...they're not accurate numbers.

I'm cool with starting at a later version (we should be able to at least know what the support is for the previous and most recent versions of the component library right?). And maybe just add a note that says something like "If possible, you should update to the most recent version for the greatest compatibility blah blah"

emclaug2: I'm fine with starting at version ^2.0.0 and adding a disclaimer.  I want to avoid people using anything lower than 3.0.0 even though it was just published.

emclaug2: Wait, this table is misleading though because 3.0.0 works with 8 and 9.


@pxblue/angular-components | @pxblue/angular-themes | @angular + @angular/material
-- | -- | --
^1.0.0 | >=4.0.0 | 8
^2.0.0 | >=5.0.0 | 8,9
^3.0.0 | >=5.1.0 | 8,9

joebochill: Looks good. I would maybe tweak the last column to use something other than a comma:
`8.x | 9.x`
`8.x - 9.x`
`^8.0.0` or `^9.0.0`

joebochill (approved):

PR #175, opened 2020-10-02
Fix display settings for infoListItem

### LINKED ISSUE #174: InfoListItem Mat Ripple Missing Divider
#### Describe the bug
Add a mat ripple onto an InfoListItem causes the divider to go missing.

Divider should work. 

https://pxblue-components.github.io/angular/?path=/story/components-info-list-item--within-a-full-list

emclaug2:
Fixes #174 

- flex-direction: column fixes the divider issue.

### DIFF
--- a/components/src/core/info-list-item/info-list-item.component.scss
+++ b/components/src/core/info-list-item/info-list-item.component.scss
@@ -6,8 +6,7 @@ $normalHeight: 72px;
 $denseHeight: 52px;
 
 .pxb-info-list-item {
-    display: flex;
-    flex-direction: column;
+    display: block;
     width: 100%;
 }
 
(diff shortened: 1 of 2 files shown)


emclaug2:
This bug was probably introduced when the host class change went in.

components/src/core/info-list-item/info-list-item.component.scss
```diff
@@ -7,6 +7,7 @@ $denseHeight: 52px;
 
 .pxb-info-list-item {
     display: flex;
+    flex-direction: column;
```
joebochill, line 10: `display: block` will give the same result with better browser support

emclaug2: You're right visually it looks correct and it's more simple, just running into a problem with the unit tests now.  

I'm gonna fix these tests then it should be good to go.

joebochill (approved):

PR #177, opened 2020-10-05
Bug/list item tag display

### LINKED ISSUE #176: ListItemTag Default Display Settings
#### Describe the bug
The default display settings for the ListItemTag take up 100% width of its container, unless the container is set to `display: flex`.

ListItemTag should not take up 100% width of container.  It should just be set to `display: block` instead of flex.

image

emclaug2:
Fixes #176 

- Change the default display settings for ListItemTag so it doesn't take up 100% width of a non-flex container.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## v3.0.1
+
+-   Fixes bug in InfoListItem where divider was working with `mat-ripple`.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+## v3.0.1
```
joebochill, line 3:
```suggestion
## v3.0.1 (Not published)
```

joebochill (approved):

PR #178, opened 2020-10-05
Feature/ng 10

emclaug2:
- Update to angular 10

#### Screenshots:
Left: Angular 8 demo
Right: Angular 10 demo
image


joebochill (commented):
Running the showcase on a fresh clone I get a compile error. Assuming this is just because we need to update the submodule pointer to point to an angular 10 version of the showcase.
image

Couple warnings in the console when running the storybook too:
image

emclaug2:
You're correct about the git submodules needing updated to use the angular 10 demo.  

This is something that I'm a little confused about.

@joebochill 
Do you know if it's possible for our current dev branch to update its submodule to point to `release/angular-8-demo` but have this branch point to a different submodule?

emclaug2:
The storybook warnings are a bit intimidating at the moment, I'll probably have to mess around with our webpack config to get something working.  The storybook app itself works without any problems.

joebochill:
> You're correct about the git submodules needing updated to use the angular 10 demo.
> 
> This is something that I'm a little confused about.
> 
> @joebochill
> Do you know if it's possible for our current dev branch to update its submodule to point to `release/angular-8-demo` but have this branch point to a different submodule?

We should be able to. The `.gitmodules` file is versioned just like any other file in git, so you should be able to change it from dev -> release/angular-8-demo in the older one, and then in the latest leave it pointed at dev.

emclaug2:
Now that the `angular-showcase-demo` is using angular 10 on dev, I've reran the circle ci build; now it should pass.

emclaug2:
`git submodule init && git submodule update` is not pulling the latest changes from dev, strange.

joebochill (approved):

emclaug2:
This PR will get merged in after current dev -> master PR gets merged in.
After this happens our component library will be migrated to use angular 10!  🎉

PR #179, opened 2020-10-05
Bug/storybook usermenu

emclaug2:
Changes: 
- Add BrowserAnimationModule to fix this story: https://pxblue-components.github.io/angular/?path=/story/components-user-menu--within-a-toolbar
- Remove unneeded import from UserMenu styles

#### Screenshots:
This is what the `within a toolbar` story looks like in master when you click on the UserMenu.  This example was missing an animation module.  Must have been accidently removed at some point. 
image

### DIFF
--- a/components/src/core/user-menu/user-menu.component.scss
+++ b/components/src/core/user-menu/user-menu.component.scss
@@ -1,4 +1,3 @@
-@import '~@angular/cdk/overlay-prebuilt.css';
 $size: 40px;
 
 /* RTL */
--- a/demos/storybook/stories/user-menu/_module.stories.ts
+++ b/demos/storybook/stories/user-menu/_module.stories.ts
@@ -74,6 +74,7 @@ storiesOf(`${COMPONENT_SECTION_NAME}/User Menu`, module)
     .addDecorator(
         moduleMetadata({
             imports: [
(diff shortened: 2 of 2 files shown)


emclaug2:
This fix will be a part of `@pxblue/angular-components`@3.0.1

After this PR gets merged in, I will make a new tag for dev and then open another PR to update dev to start using angular 10.

joebochill:
This is just a change to the storybook — we shouldn't need a new version of the library to get this in...we can just make this PR into dev and then another one from this feature branch into master.

joebochill (approved):

emclaug2:
> This is just a change to the storybook — we shouldn't need a new version of the library to get this in...we can just make this PR into dev and then another one from this feature branch into master.

Yeah you're right, we still have some bug fixes in v3.0.1 that'll have to be published at some point.  I'm going to keep my eyes open to see if I can find anymore issues before v3.0.1 is published.

huayunh:
@emclaug2  If you need something else to fix for 3.0.1: 
https://github.com/pxblue/angular-component-library/pull/167#pullrequestreview-497969935

(Same for all the other component libraries, I don't remember if we have a story for this in the backlog)

PR #181, opened 2020-10-06
Use angular 8 showcase for angular 8 version of comp library

emclaug2:
- Use `release/angular-8-demo` for our angular 8 component libary

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "demos/showcase"]
 	path = demos/showcase
 	url = https://github.com/pxblue/angular-showcase-demo
-	branch = dev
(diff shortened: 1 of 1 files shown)


joebochill (approved):
Looks good...dev branch of the component library is still building with 8 right? Then we switch this back when we make the NG10 the standard in dev?

emclaug2:
> Looks good...dev branch of the component library is still building with 8 right? Then we switch this back when we make the NG10 the standard in dev?

Yep the feature/ng-10 branch will switch us back to using dev.  I'm going to make a `release/3.X.X` branch right now that uses `release/angular-8-demo` .

PR #189, opened 2020-10-21
Feature/drawer nav item hidden

### LINKED ISSUE #185: Hidden Nav Items
User should be able to conditionally hide nav items that are defined in the items list.

All items are rendered. No option to hide them.

Adopter Request.

#### Additional Context
The primary use case for this is role-based routes. Users should not be able to see/access some of the admin only pages of the application. There is an equivalent issue already for React and React Native.
https://github.com/pxblue/react-component-library/issues/197
https://github.com/pxblue/react-native-component-library/issues/83

emclaug2:
Fixes #185 

- Add `isVisible` prop to nav items.  Default value is `true`.

### DIFF
--- a/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.ts
+++ b/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.ts
@@ -36,71 +36,73 @@ export type ActiveItemBackgroundShape = 'round' | 'square';
     encapsulation: ViewEncapsulation.None,
     styleUrls: ['./drawer-nav-item.component.scss'],
     template: `
-        <ng-template #navIcon><ng-content select="[pxb-icon]"></ng-content></ng-template>
-        <div class="pxb-drawer-nav-item-content" [class.pxb-drawer-nav-item-active]="selected">
-            <div
-                matRipple
-                [class.pxb-drawer-nav-item-active-highlight]="selected"
-                [class.round]="activeItemBackgroundShape === 'round' && isOpen() && !isRail()"
-                [class.square]="activeItemBackgroundShape === 'square' || !isOpen() || isRail()"
-            ></div>
-            <pxb-info-list-item
-                *ngIf="!isRail()"
-                (click)="selectItem($event)"
-                [dense]="true"
-                [statusColor]="statusColor"
-                [chevron]="chevron && isOpen()"
-                [hidePadding]="hidePadding"
-                [divider]="divider ? 'full' : undefined"
-                [class.no-icon-closed]="isEmpty(iconEl) && !isOpen()"
-            >
-                <ng-container pxb-icon #icon>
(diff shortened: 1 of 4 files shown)


huayunh:
I think "isHidden" or just "hidden" might be more intuitive than "isVisible"

I assume that everytime you introduce a new prop/attribute, it is a derivative from the default use case. 

Since the default is visible, if the user wants to modify it from the default, they will use the"hidden" attribute

emclaug2:
> I think "isHidden" or just "hidden" might be more intuitive than "isVisible"
> 
> I assume that everytime you introduce a new prop/attribute, it is a derivative from the default use case.
> 
> Since the default is visible, if the user wants to modify it from the default, they will use the"hidden" attribute

I like this approach; I've updated the prop name to `isHidden`.

components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.ts
```diff
@@ -117,6 +120,7 @@ export class DrawerNavItemComponent extends StateListener implements Omit<Drawer
     @Input() statusColor: string;
     @Input() subtitle: string;
     @Input() title: string;
+    @Input() isHidden = false;
```
joebochill, line 123: correct me if I'm wrong, but I don't think we typically use the `is_` prefix for attributes/props/Inputs. I would go for `hidden` instead if it were me.

docs/Drawer.md
```diff
@@ -304,6 +304,7 @@ Parent element (`<pxb-drawer-nav-item>`) attributes:
 | divider                   | Sets whether to show divider          | `boolean`            | no       | true    |
 | expanded                  | Sets whether to show nested nav items | `boolean`            | no       | false   |
 | hidePadding               | Sets whether to show/hide padding     | `boolean`            | no       |         |
+| isHidden                  | Sets whether to show the nav item     | `boolean`            | no       | false   |
```
joebochill, line 307:
```suggestion
| hidden                  | Sets whether to hide the nav item     | `boolean`            | no       | false   |
```

joebochill (approved):

joebochill:
Unit test to verify that it works?

joebochill:
Anything we want to show in Storybook?

emclaug2:
> Unit test to verify that it works?

I'll add a unit test to show that it works; I don't think this change is big enough to warrant it's own story in storybook.  I've considered adding a knob to the full config story but found a new knob a bit confusing.    I think the api docs should be enough for people to know how to use this new prop.

components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.ts
emclaug2: I've removed the `is` part of the `isHidden`

PR #191, opened 2020-10-26
Add DotStepper component

emclaug2:
- Add Dot Stepper component to angular component library.
- This includes new component, tests, storybook stories, and docs.

The stepper has a few default colors that will be overwritten by a theme upgrade.  The storybook deploy below uses the new theme; if you checkout this branch and run locally, the colors in the stepper will be different. 

Active color: blue[500]
Inactive color: gray[200]

https://pxblue-angular-library.web.app/?path=/info/components-dot-stepper--readme


docs/DotStepper.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Classes
+
+Each PX Blue component has classes which can be used to override component styles:
+
+| Name                           | Description                                             |
+| ------------------------------ | ------------------------------------------------------- |
+| pxb-dot-stepper                 | Styles applied to the tag                              |
+| pxb-dot-stepper-dot            | Styles applied to each dot in the stepper               |
+| pxb-dot-stepper-active         | Styles applied to the dot representing the active step  |
```
huayunh, line 46: We can add two extra classes, one for the "visited" dots, and one for "unvisited" / "future" dots

huayunh:
Usually there are some width to the modal card. Make sure your buttons and progress dots are properly flexed and aligned

image

I suggest wrap another flex=1 div container around the component

emclaug2:
`activeIndexId` is kind of a weird name because it implies the active identifier can be something other than a number.  Thoughts on changing this prop name to `activeStepNumber` or `activeIndex`?

huayunh:
I vote for `activeIndex`

emclaug2:
> Usually there are some width to the modal card. Make sure your buttons and progress dots are properly flexed and aligned
> 
> image
> 
> I suggest wrap another flex=1 div container around the component

The stepper is going to have to be placed in a container that typically has an order of button, stepper, button.  That parent container is what is going to have the flex settings applied so the content won't be pushed into a corner.

emclaug2:
I've added in the classes for visited and unvisited dots.  This'll be helpful for end-users to apply conditional styles if they choose.

docs/DotStepper.md
emclaug2: Good suggestion, I've added these two classes.

components/src/core/dot-stepper/dot-stepper.component.scss
```diff
@@ -0,0 +1,17 @@
+.pxb-dot-stepper {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+}
+
+.pxb-dot-stepper-dot {
+    height: 8px;
+    width: 8px;
+    background-color: gray;
```
huayunh, line 10: Why is this color style here? Do they belong to pxb themes instead?

emclaug2: I've added in temporary styles because otherwise this component will not render anything.  I think it would be a good idea to have some sort of default colors in the event someone is not using our themes and expects to visually see something.

huayunh:
> The stepper is going to have to be placed in a container that typically has an order of button, stepper, button. That parent container is what needs display: flex and probably space-between applied to it.

I personally think the "space between" can be moved to the dot stepper as "flex:1" instead — Just one less thing for our adopters to worry about. I cannot think of a use case where the flex:1 might get in the way. Suggestions @joebochill ?

emclaug2:
> > The stepper is going to have to be placed in a container that typically has an order of button, stepper, button. That parent container is what needs display: flex and probably space-between applied to it.
> 
> I personally think the "space between" can be moved to the dot stepper as "flex:1" instead — Just one less thing for our adopters to worry about. I cannot think of a use case where the flex:1 might get in the way. Suggestions @joebochill ?

If our adopters need to add more/less space between the dots they always have access to override the css.  If we want to support a dynamic spacing option maybe we could add a variant to our dot stepper that dynamically spaces the dots.

joebochill:
I'm inclined to say that the back and next buttons should be part of the stepper API. The whole 'footer' should be the component, not just the dots.

We should be matching as closely as possible with the MUI MobileStepper (https://material-ui.com/api/mobile-stepper/), including `@Input` names and the ability to do text, progress, or dots. I would also suggest naming the component MobileStepper to be consistent.

emclaug2:
> I'm inclined to say that the back and next buttons should be part of the stepper API. The whole 'footer' should be the component, not just the dots.
> 
> We should be matching as closely as possible with the MUI MobileStepper (https://material-ui.com/components/steppers/#mobile-stepper)(https://material-ui.com/api/mobile-stepper/), including `@Input` names and the ability to do text, progress, or dots. I would also suggest naming the component MobileStepper to be consistent.

For the sake of making this PR doable in a day, I think it'll be best to limit the first version of this component to only support the Dots and Numbers.  I'll do the name refactor too and enhance the component to include a default BACK and NEXT button.

I think the buttons style should also be something that users might want to be able to customize.  I'll think about how we can achieve that today.

joebochill:
> For the sake of making this PR doable in a day, I think it'll be best to limit the first version of this component to only support the Dots and Numbers. I'll do the name refactor too and enhance the component to include a default BACK and NEXT button.
> 
> I think the buttons style should also be something that users might want to be able to customize. I'll think about how we can achieve that today.

I think we can just do the numbers for now (as long as we have enhancements to add the other two formats) — they should be pretty easy if we have the number of steps and the current step to swap out a progress bar for the dots. The buttons can just be passed in as ng-content. That's how the React implementation works. We jsut have a dedicated place for them in the footer so that we can control the flexing like Huayun was talking about. They could pass in a regular button, an icon button, or whatever.

emclaug2:
I've updated the name of the `DotStepperComponent` to `MobileStepperComponent` and introduced 3 new variants to it - `dots`, `text`, and `progress`.

dots - shows dot indicator
text - shows text progress ( e.g. 1/ 5 ) 
progress - show progress bar

I've also updated the component to include default `next` and `back` buttons.  There are two new event emitters added so users can respond to button presses.

I'm still considering adding `ng-content` button overrides in the event that users want to provide their own buttons text & style.  Maybe this is a future enhancement. 

Updates can be found here: https://pxblue-angular-library.web.app/?path=/info/components-mobile-stepper--readme

huayunh:
> I'm still considering adding ng-content button overrides in the event that users want to provide their own buttons text & style. Maybe this is a future enhancement.

Looks like that is how the React variant is implemented, yes. I will take a look at this PR later.

emclaug2:
I'm currently working on making the `next` and `back` button be customizable via ng-content.  I think this is one of the last things this component needs. 

I'm going to add a new storybook story demoing this feature. 

Anything else I'm missing here?

joebochill:
> I'm currently working on making the `next` and `back` button be customizable via ng-content. I think this is one of the last things this component needs.
> 
> I'm going to add a new storybook story demoing this feature.
> 
> Anything else I'm missing here?

I think it's looking good. I might even avoid the default buttons altogether and just take them exclusively through ng-content (that's how react does it). There's not really an advantage of us controlling them since the current step is passed in through inputs anyway. The dot colors also look a little off on the storybook, but it might just be me. 

image

joebochill:
> The stepper has a few default colors that will be overwritten by a theme upgrade. The storybook deploy below uses the new theme; if you checkout this branch and run locally, the colors in the stepper will be different.

the deployed storybook and the local running storybook both appear to be using the incorrect colors ('blue' and 'gray'). Do we have a PR for themes with the necessary update?

components/src/core/mobile-stepper/mobile-stepper.component.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+        fixture.detectChanges();
+        void expect(component).toBeTruthy();
+    });
+
+    it('should render the correct number of steps', () => {
+        fixture.detectChanges();
+        const dots = fixture.nativeElement.querySelectorAll('.pxb-mobile-stepper-dot');
+        void expect(dots.length).toBe(4);
+    });
+
```
joebochill, line 46: can we test the correct active styles are applied to the activeStep?

components/src/core/mobile-stepper/mobile-stepper.component.ts
```diff
@@ -0,0 +1,33 @@
...
+    styleUrls: ['./mobile-stepper.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.None,
+    host: {
+        class: 'pxb-mobile-stepper',
+    },
+})
+export class MobileStepperComponent implements OnChanges {
+    @Input() steps: number;
+    @Input() activeIndex: number;
```
joebochill, line 18: MUI calls this `activeStep`

docs/MobileStepper.md
```diff
@@ -0,0 +1,73 @@
+# Mobile Stepper
+
+The `<pxb-mobile-stepper>` is used to minimally display progress when completing a workflow that requires multiple steps.  There are three types of progress indicators supported, `dots`, `text`, and `progress`.
```
joebochill, line 3:
```suggestion
The `<pxb-mobile-stepper>` is used to minimally display progress when completing a workflow that requires multiple steps.  There are three types of progress indicators supported, `dots` (default), `text`, and `progress`. 
```

docs/MobileStepper.md
```diff
@@ -0,0 +1,73 @@
...
+imports: [
+    MobileStepperModule
+],
+```
+
+```html
+// your-component.ts
+activeIndex = 0;
+steps = 4;
+```
```
joebochill, line 24: do we need to show these variable definitions? Would it be cleaner just to leave them out?

emclaug2: For the code example to work, our end users will have to define `activeIndex` as a variable in their component.  It's their responsibility to keep track of which dot number is active.  We could leave out the `dots` variable though.

components/src/core/mobile-stepper/mobile-stepper.component.ts
emclaug2: Ah for consistency I'll update this.

components/src/core/mobile-stepper/mobile-stepper.component.spec.ts
emclaug2: The class-name convention story tests that the `pxb-mobile-stepper-dot-active` class is applied.  Since the colors are provided from an external theme we won't be able to test for specific color, just class presense.

emclaug2:
> > The stepper has a few default colors that will be overwritten by a theme upgrade. The storybook deploy below uses the new theme; if you checkout this branch and run locally, the colors in the stepper will be different.
> 
> the deployed storybook and the local running storybook both appear to be using the incorrect colors ('blue' and 'gray'). Do we have a PR for themes with the necessary update?


I've just updated the firebase deployment to use an updated theme that adds colors to the MobileStepper!   I'm going to open another PR now for these theme changes.

https://pxblue-angular-library.web.app

components/src/core/mobile-stepper/mobile-stepper.component.ts
emclaug2: Updated

docs/MobileStepper.md
joebochill: it's not a big deal, I was just wondering if it made sense to simplify. We could hardcode numbers in the binding instead of using variables, but I wouldn't worry about it.

components/src/core/mobile-stepper/mobile-stepper.component.spec.ts
joebochill: could we check that the class is applied on the 0th item?

joebochill (approved):

PR #193, opened 2020-11-12
Drawer Shadow and EnableSelectionHierarchy

emclaug2:
- Add a 4px shadow to the Drawer by default.
- When `Drawer` `sideBorder` prop is `false`, shadow is gone.
- Add `enableSelectionHierarchy` prop to `Drawer`, update storybook stories and docs.

#### Screenshots:
image

### DIFF
--- a/components/src/core/drawer/drawer-layout/drawer-layout.component.scss
+++ b/components/src/core/drawer/drawer-layout/drawer-layout.component.scss
@@ -8,6 +8,9 @@ $transition: 300ms cubic-bezier(0.4, 0, 0.2, 1);
 /* RTL */
 [dir='rtl'] .pxb-drawer-layout-content .mat-drawer-side {
     border-left: 0; // Border-left is provided by the Drawer component.
+    .pxb-drawer-layout-sidenav {
+        box-shadow: 0 0 -4px 0 #424e54;
+    }
 }
 
 .pxb-drawer-layout-content {
@@ -31,6 +34,10 @@ $transition: 300ms cubic-bezier(0.4, 0, 0.2, 1);
         &.pxb-drawer-collapse {
             width: 100%;
         }
+    }
+
+    .pxb-drawer-layout-sidenav {
+        box-shadow: 0 0 4px 0 #424e54;
     }
 
     .pxb-drawer-layout-sidenav,
--- a/components/src/core/drawer/drawer-layout/drawer-layout.component.scss
+++ b/components/src/core/drawer/drawer-layout/drawer-layout.component.scss
@@ -36,7 +36,7 @@ $transition: 300ms cubic-bezier(0.4, 0, 0.2, 1);
         }
     }
 
-    .pxb-drawer-layout-sidenav.pxb-drawer-layout-side-border  {
+    .pxb-drawer-layout-sidenav.pxb-drawer-layout-side-border {
         box-shadow: 0 0 4px 0 #424e54;
     }
 
--- a/components/src/core/drawer/drawer-layout/drawer-layout.component.scss
+++ b/components/src/core/drawer/drawer-layout/drawer-layout.component.scss
@@ -8,7 +8,7 @@ $transition: 300ms cubic-bezier(0.4, 0, 0.2, 1);
 /* RTL */
 [dir='rtl'] .pxb-drawer-layout-content .mat-drawer-side {
     border-left: 0; // Border-left is provided by the Drawer component.
-    .pxb-drawer-layout-sidenav.pxb-drawer-layout-side-border {
+    .pxb-drawer-layout-sidenav.pxb-drawer-layout-shadow {
         box-shadow: 0 0 -4px 0 #424e54;
     }
 }
@@ -36,7 +36,7 @@ $transition: 300ms cubic-bezier(0.4, 0, 0.2, 1);
         }
     }
 
-    .pxb-drawer-layout-sidenav.pxb-drawer-layout-side-border {
+    .pxb-drawer-layout-sidenav.pxb-drawer-layout-shadow {
         box-shadow: 0 0 4px 0 #424e54;
     }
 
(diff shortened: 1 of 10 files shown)


emclaug2:
I added this border in two places because the Drawer's border was being lost when placed inside of a PXB Drawer Layout.   So this border is applied when a Drawer exists in or out of a PXB DrawerLayout.

huayunh (changes requested):
Can you add `git submodule update --remote` to package json?

----
image
The showcase appears to be using the unelevated version? And the parent items are not 600 font weight?

The discussion today was to turn on the selection hierarchy to true by default, and allow the user to turn it off 

Also I feel like the term "selection hierarchy" might not be descriptive enough. I might do "bolden parent nav items" or its alike... open to suggestions.

emclaug2:
```
| sideBorder               | Toggle a side border instead of shadow                        | `boolean` | no       | false   |
```

Is this description accurate?  So the expectation is that a Drawer will always come with a vertical dividing line in the form of  a shadow or an actual border?

emclaug2:
> Also I feel like the term "selection hierarchy" might not be descriptive enough. I might do "bolden parent nav items" or its alike... open to suggestions.

Here's a couple I've got; let's wait for @joebochill to chime in on this come Monday.  

`enableSelectedParentHighlight`
`enableTreeSelection`
`boldSelectedItemTree`
`boldSelectedItemParents`
`enableSelectionHierarchy`

Also the submodule command was added to the package.json now and this selection hierarchy is enabled by default now.

huayunh:
> sideBorder

I don't know what's the naming convention on the ng side. Maybe `outlined`? When off, then the elevation will be used. 

The user should not use borders in conjunction with elevation.

components/src/core/drawer/drawer-layout/drawer-layout.component.scss
```diff
@@ -8,6 +8,9 @@ $transition: 300ms cubic-bezier(0.4, 0, 0.2, 1);
 /* RTL */
 [dir='rtl'] .pxb-drawer-layout-content .mat-drawer-side {
     border-left: 0; // Border-left is provided by the Drawer component.
+    .pxb-drawer-layout-sidenav.pxb-drawer-layout-side-border {
+        box-shadow: 0 0 -4px 0 #424e54;
+    }
```
joebochill, line 13: should these colors come from the theme? Or at least the Colors package?

components/src/core/drawer/drawer.component.ts
```diff
@@ -41,6 +41,7 @@ export class DrawerComponent extends StateListener implements OnInit, OnChanges
     @Input() open: boolean;
     @Input() condensed = false;
     @Input() sideBorder = false;
+    @Input() enableSelectionHierarchy = true;
```
joebochill, line 44: can we make this more descriptive? Like `disableActiveItemParentStyles` or something like that? Should is be disable____ instead of enable______ since the default is on?

components/src/core/drawer/drawer.component.scss
```diff
@@ -12,6 +16,7 @@ $transition: width 300ms cubic-bezier(0.4, 0, 0.2, 1) 0ms;
     border-right: 0;
     &.pxb-drawer-side-border {
         border-left: 1px solid rgba(0, 0, 0, 0.12);
```
joebochill, line 18: why is this a border-left?

joebochill (changes requested):
border and shadow should be opposites. If the sideborder is on, there is no shadow. If sideborder is off, then there is a shadow.

components/src/core/drawer/drawer-layout/drawer-layout.component.scss
emclaug2: Since the pxb-angular-comp library doesn't have a peerDependency on our colors package we can add an override for this color in our themes package.  This can be another instance of having a default color baked into our library, which we just override with our pxb & pxb-dark theme.

components/src/core/drawer/drawer.component.ts
emclaug2: I'm happy with the name `disableActiveItemParentStyles `, I'll update accordingly.

emclaug2: This is updated now.

components/src/core/drawer/drawer.component.scss
emclaug2: These are the RTL styles.   The border-right style is down below, line 35.

emclaug2:
> border and shadow should be opposites. If the sideborder is on, there is no shadow. If sideborder is off, then there is a shadow.

This should be the last pending item for this PR; working on this right now.

For the border-color, I've made an issue for our themes repo, here: https://github.com/pxblue/themes/issues/117

docs/Drawer.md
```diff
@@ -23,12 +23,12 @@ Parent element (`<pxb-drawer>`) attributes:
...
-| ------------------------ | ------------------------------------------------------------- | --------- | -------- | ------- |
-| condensed                | Skinny view for `rail` variant                                | `boolean` | no       | false   |
-| enableSelectionHierarchy | Bold nav item group title when a nested item is selected      | `boolean` | no       | true    |
-| open                     | State for the drawer                                          | `boolean` | yes      |         |
-| sideBorder               | Toggle a side border instead of shadow                        | `boolean` | no       | false   |
+| @Input                        | Description                                                   | Type      | Required | Default |
+| ----------------------------- | ------------------------------------------------------------- | --------- | -------- | ------- |
+| condensed                     | Skinny view for `rail` variant                                | `boolean` | no       | false   |
+| disableActiveItemParentStyles | Bold nav item group title when a nested item is selected      | `boolean` | no       | false   |
```
joebochill, line 29:
```suggestion
| disableActiveItemParentStyles | If true, NavItems will not have a bold title when a child NavItem is selected | `boolean` | no       | false   |
```

joebochill (approved):

PR #194, opened 2020-12-01
Hero margin 5px -> 8px

huayunh:
Same fix as https://github.com/pxblue/react-component-library/pull/206


joebochill:
Same comment about 12 vs 8px

joebochill (approved):

PR #200, opened 2020-12-21
Fix Condensed Rail Tooltip Issue Storybook

### LINKED ISSUE #199: Storybook Nav Item ToolTip Issue
#### Describe the bug
Storybook is having change-detection issues whenever we switch from `condensed=false` to `condensed=true`.

This shouldn't be happening: 
nav-item-tooltip-issue

1. Go to https://pxblue-components.github.io/angular/?path=/story/components-drawer--with-a-nav-rail
2. Make drawer condensed
3. Hover Nav Items and observe tooltips

#### Environment

Windows, Chrome		

#### Anything else to add?
I blame storybook.

emclaug2:
Fixes #199

- If `condensed=true` is the default, storybook no longer has issues with creating/destroying tooltips.

### DIFF
--- a/demos/storybook/stories/drawer/with-nav-rail.stories.ts
+++ b/demos/storybook/stories/drawer/with-nav-rail.stories.ts
@@ -57,7 +57,7 @@ export const withNavRail = (): any => ({
     props: {
         navItems: navItems,
         headerImg: headerImg,
-        condensed: boolean('condensed', false),
+        condensed: boolean('condensed', true),
         divider: boolean('divider', false),
     },
 });


daileytj (approved):

emclaug2:
I'm doing a sanity check just to make sure that the original issue is contained only to storybook.  Storybook has had a similar issue in the past already.

emclaug2:
> I'm doing a sanity check just to make sure that the original issue is contained only to storybook. Storybook has had a similar issue in the past already.

No issues found 👍

Issue #202, opened 2021-01-06
Incorrect shadow applied on Drawer


joebochill:
#### Describe the bug
The shadow on the Drawer component does not match the design in Figma. We have a single box shadow defined currently: 

#### What is the expected behavior?
Component should be using the styles for 4dp shadow (elevation = 4). These should be coming from the mat-elevation class or mixin, which includes a blend of box shadows

#### What are the steps to reproduce?
1. Go to...
2. Click on...

#### Screenshots or links to minimum reproduction example
Current definition:
![image](https://user-images.githubusercontent.com/29152776/103773878-ba8a2a00-4ff9-11eb-9184-7399e7ff27f2.png)

Should look more like:
![image](https://user-images.githubusercontent.com/29152776/103773922-c83faf80-4ff9-11eb-9323-d73594281566.png)
but with the correct black[500] color.

Currently:
![image](https://user-images.githubusercontent.com/29152776/103774011-e9a09b80-4ff9-11eb-9105-1cf49e5139bb.png)

Should look like:
![image](https://user-images.githubusercontent.com/29152776/103774054-f91fe480-4ff9-11eb-83a9-dfec2d6dc418.png)

		
#### Anything else to add?
This is probably something that needs to be adjusted in the component library and the theme (to get the right colors).

emclaug2:
#210 fixes this.

PR #210, opened 2021-01-18
Audit changes and use latest theme

emclaug2:

Use latest theme package @^5.2.0

Channel Value Changes:
-  Change `full config` icon color to red[500].

Empty State Changes:
- Made action button outlined variant in both button examples

InfoListItem Changes:
- Font and Bg color changes to be more app-reaslistic.

Scorecard Changes:
- Replace all red[500] bg color headers with blue[500]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 -   Added new property `openOnHover` to `<pxb-drawer>`.
 -   Added lighter font weight to `<pxb-info-list-item>` placed inside `<pxb-user-menu>`.
+-   Added a shadow to the `<pxb-user-menu>` when opened.
 
 ### Fixed
 
@@ -14,6 +15,7 @@
 -   Fixed misalignment of `<pxb-info-list-item>`'s right content.
 -   Fixed RTL styles in `<pxb-dropdown-toolbar>`.
 -   Fixed uneven vertical alignment of icon and text in `<pxb-empty-state>` action buttons. 
+-   Fixed `<pxb-user-menu>` menu header avatar offset.
 
 ### Removed
 -   Removed automatic RTL flip logic from user-provided icons to `<pxb-info-list-item>`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,10 @@
 -   Fixed bug in `<pxb-mobile-stepper>` which makes component span 100% of parent width.
 -   Fixed header height bug which affected `<pxb-drawer-header>` in Safari.
 -   Fixed misalignment of `<pxb-info-list-item>`'s right content.
+-   Fixed RTL styles in `<pxb-dropdown-toolbar>`.
 
 ### Removed
+
 -   Removed icon inversion logic from `<pxb-info-list-item>` when viewing right-to-left directionality.
 
 ## v4.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 -   Fixed header height bug which affected `<pxb-drawer-header>` in Safari.
 -   Fixed misalignment of `<pxb-info-list-item>`'s right content.
 -   Fixed RTL styles in `<pxb-dropdown-toolbar>`.
+-   Fixed uneven vertical alignment of icon and text in `<pxb-empty-state>` action buttons. 
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 -   Added new property `openOnHover` to `<pxb-drawer>`.
+-   Added lighter font weight to `<pxb-info-list-item>` placed inside `<pxb-user-menu>`.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 -   Fixed RTL styles in `<pxb-dropdown-toolbar>`.
 -   Fixed uneven vertical alignment of icon and text in `<pxb-empty-state>` action buttons. 
 -   Fixed `<pxb-user-menu>` menu header avatar offset.
+-   Fixed `<pxb-drawer>` `sideBorder` prop to use `mat-elevation` class.  
 
 ### Removed
 -   Removed automatic RTL flip logic from user-provided icons to `<pxb-info-list-item>`.
--- a/components/src/core/drawer/drawer.component.scss
+++ b/components/src/core/drawer/drawer.component.scss
@@ -18,7 +18,7 @@ $transition: width 300ms cubic-bezier(0.4, 0, 0.2, 1) 0ms;
         border-left: 1px solid rgba(0, 0, 0, 0.12);
     }
     &:not(.pxb-drawer-side-border) {
-        box-shadow: 0 0 -4px 0 #424e54;
+        border-left: 0;
     }
     &.temp-variant {
         border-left: 0;
@@ -36,7 +36,7 @@ $transition: width 300ms cubic-bezier(0.4, 0, 0.2, 1) 0ms;
         border-right: 1px solid rgba(0, 0, 0, 0.12);
     }
     &:not(.pxb-drawer-side-border) {
-        box-shadow: 0 0 4px 0 #424e54;
+        border-right: 0;
     }
     &.pxb-drawer-temp-variant {
         border-right: 0;
--- a/components/src/core/drawer/drawer.component.ts
+++ b/components/src/core/drawer/drawer.component.ts
@@ -19,6 +19,7 @@ import { Subscription } from 'rxjs';
         <div
             class="pxb-drawer-content"
             [class.pxb-drawer-side-border]="sideBorder"
+            [class.mat-elevation-z4]="!sideBorder"
             [class.pxb-drawer-condensed-rail]="condensed"
             [class.pxb-drawer-collapse]="!isOpen()"
             [class.pxb-drawer-temp-variant]="isTemporaryVariant()"
--- a/components/src/core/dropdown-toolbar/dropdown-toolbar.component.scss
+++ b/components/src/core/dropdown-toolbar/dropdown-toolbar.component.scss
@@ -1,7 +1,7 @@
 [dir='rtl'] .pxb-dropdown-toolbar-content {
-    .pxb-dropdown-toolbar-icon-wrapper {
+    .pxb-dropdown-toolbar-icon-wrapper > * {
         margin-right: -8px;
-        margin-left: 16px;
+        margin-left: 24px;
     }
     .pxb-dropdown-toolbar-text-content-container {
         padding-right: 0;
--- a/components/src/core/empty-state/empty-state.component.scss
+++ b/components/src/core/empty-state/empty-state.component.scss
@@ -35,4 +35,9 @@
             text-align: center;
         }
     }
+
+    .pxb-empty-state-actions-wrapper .mat-button-wrapper {
+        display: flex;
+        align-items: center;
+    }
 }
--- a/components/src/core/user-menu/user-menu.component.scss
+++ b/components/src/core/user-menu/user-menu.component.scss
@@ -1,13 +1,5 @@
 $size: 40px;
 
-/* RTL */
-[dir='rtl'] {
-    .pxb-user-menu-header-avatar {
-        margin-left: 0;
-        margin-right: 4px;
-    }
-}
-
 .pxb-user-menu {
     display: flex;
 }
@@ -36,10 +28,6 @@ $size: 40px;
 
 mat-card.pxb-user-menu-overlay {
     padding: 0;
-}
-
-.pxb-user-menu-header-avatar {
-    margin-left: 4px;
 }
 
 .pxb-user-menu-text-avatar {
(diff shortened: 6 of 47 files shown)


joebochill:
Same comments as a dropped on the React PR (hey, the demos are the same! 😄 ), plus:

- Drawer Rail example should default to _not_ condensed
- ILI full list...leaf is not an icon of temperature

emclaug2:
I'm working on matching React's comments right now 🗡️ 

>>Drawer Rail example should default to not condensed

This was changed to prevent this from happening: 
nav-item-tooltip-issue

joebochill:
> I'm working on matching React's comments right now 🗡️
> 
> > > Drawer Rail example should default to not condensed
> 
> This was changed to prevent this from happening:
> 
> nav-item-tooltip-issue
> 
> I blame storybook because I was never able to reproduce this outside of storybook.

Lol what the heck

joebochill:
> > > Drawer Rail example should default to not condensed
> 
> This was changed to prevent this from happening:
> 
> I blame storybook because I was never able to reproduce this outside of storybook.

Lol what the heck

emclaug2:
This PR is ready for another round of review; there are some outstanding RTL issues in the UserMenu and DropdownToolbar that I'm working on right now.

I think the RTL stuff in our overlays are a bit funky right now.

emclaug2:
Here's a summary of how our components currently handle icon-inversion in RTL.    At some point all components were set up to do the inversion but this logic was removed.  If this is looking fine then I'm going to manually add those RTL inversions to all required storybook stories. 

`<pxb-hero>`: no-flip
`<pxb-channel-value>`: no-flip
`<pxb-drawer>`: flip
`<pxb-drawer> (rail)` no-flip
`<pxb-info-list-item>`: flip
`<pxb-empty-state>`: no-flip
`<pxb-score-card>`: no-flip

huayunh:
Yeah that was my fault — only icons indicating directionalities should be flipped. 

See https://px-blue.github.io/patterns/internationalization#bidirectionality--right-to-left-(rtl)-support for what to flip

huayunh:
So essentially when a user passes in a custom icon, it is the user's job to flip icons, not that of the component library. 

Icons that are part of the component library should be flipped on our behalf. For example, the chevron on an Info List Item

emclaug2:
> So essentially when a user passes in a custom icon, it is the user's job to flip icons, not that of the component library.
> 
> Icons that are part of the component library should be flipped on our behalf. For example, the chevron on an Info List Item

I'm okay with this idea as long as are consistent; I think we should take out the icon flipping logic from the InfoListItem component (that'll remove it from Drawer too).

joebochill:
> So essentially when a user passes in a custom icon, it is the user's job to flip icons, not that of the component library.
> 
> Icons that are part of the component library should be flipped on our behalf. For example, the chevron on an Info List Item

This is correct and it's how we are doing it in react right now. We can't auto-flip icons that are passed in by the user because we don't know what they are (they may or may not need to be flipped). The one's we use internally, we can flip because we know what they are. I believe there might be an older issue already logged about this.

emclaug2:
This PR also fixes a RTL menu-icon spacing bug in the DropdownToolbar.  

Stories should match between React and Angular pretty closely now.  Biggest differences are `full config` and `nested nav items` stories for the Drawer.  

React's nested nav item story shows an example of 4 levels of nav items which isn't something we should be encouraging users to do, so I left Angular the way it was.

emclaug2:
Hosted here: https://pxblue-angular-library.web.app/

emclaug2:
This PR also has the InfoListItem RTL inversion logic removed (which is in a separate PR).

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-channel-value--with-full-config
Flip icon

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-drawer--with-basic-usage
Identity management — use the solid icon, not the outlined version

huayunh:
Same for https://pxblue-angular-library.web.app/?path=/story/components-drawer--with-nested-nav-items

huayunh:
I suppose this is a theme problem but the selected nav item should turn blue (for both icons and labels)
image

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-empty-state--with-full-config 
The size of the button differs between angular and React... pick your favorite one :P

huayunh:
Make a story or log a bug to make Mobile Stepper left-to-right friendly

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-user-menu--with-basic-usage
image
* React is using font weight 400 here
* Flip the logout icon

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-user-menu--within-a-toolbar
version number ought to be pinned to the other side
image

huayunh:
Is ng dark theme still due for an update at some point? Right now the dropdown toolbar shares the same color as its background under dark theme
image

huayunh:
button text and icon are not perfectly aligned

image

Need a 
```css
{
    display: flex;
    align-items: center;
}
```
somewhere for the `.mat-button-wrapper`

And change icon font size to be 20px instead of 24px

joebochill:
> Is ng dark theme still due for an update at some point? Right now the dropdown toolbar shares the same color as its background under dark theme

Yeah, we have an upcoming story to address dark theme changes once we do an audit.

emclaug2:
> https://pxblue-angular-library.web.app/?path=/story/components-empty-state--with-full-config
> The size of the button differs between angular and React... pick your favorite one :P

I've remove React's `small` prop from its button.

emclaug2:
This is updated now: https://pxblue-angular-library.web.app/?path=/story/*

Empty State add icon is 20px now.
Icon and Text in EmptyState action buttons content is vertically centered. (component change)
UserMenu icons are RTL-inverted.
UserMenu list item font weight is 400 now. (component change)

huayunh:
image

Do we want to pick a circular icon (like `voltage_circled_outline`) so that the overflowing icon issue is less obvious?

huayunh:
Use a better subheader component here
image

And change the icon of the identity management to be solid

huayunh:
If you want to go fancy on drawer subheader... instead of doing what React is doing, you can try something like this
image
or do an organization switcher
image
that would trigger a dropdown to select an org / a location

emclaug2:
> If you want to go fancy on drawer subheader... instead of doing what React is doing, you can try something like this
> image
> or do an organization switcher
> image
> that would trigger a dropdown to select an org / a location

I think this is better as a separate issue since it'd have to be done for angular and react.

emclaug2:
Ok latest changes are live here:

- Drawer person icon should be filled now.
- Barley icon replaced with better, rounder icon
- Drawer subheader looks better (matching react)

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-drawer--with-full-config
Update the menu items so that they make more sense (match previous stories)

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-info-list-item--with-full-config
image
Left icon not center-aligned

Same for the first info list item of https://pxblue-angular-library.web.app/?path=/story/components-info-list-item--within-a-full-list
image
All the other icons of ILI appear fine to me

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-score-card--with-full-config

image
Flip the list icon, or replace it with `favorite` or `bookmark` (which are two more commonly used icons for scorecard)

huayunh:
Don't know if this is a storybook issue or ACL issue —
image

Step to reproduce: 
1. Go to https://pxblue-angular-library.web.app/?path=/story/components-user-menu--within-a-toolbar  (Or any user menu story)
2. Without touching anything else, toggle the i18n on so that it reads RTL
3. Now click the avatar. Observe what you see in the screenshot

Note that this would not occur if you enter a story with i18n already turned on (i.e., turn on i18n and switch between different stories)
This would not happen either if you click the avatar first, and then toggle i18n

huayunh:
Shrink the user menu footer height in the within a toolbar story.
React does 40px tall
image

huayunh:
I noticed that there is a 4px left margin on `.pxb-user-menu-header-avatar`. This caused misalignment of several UI elements: 

image

huayunh:
Oh and can we give the user menu an 8dp shadow?

emclaug2:
> https://pxblue-angular-library.web.app/?path=/story/components-score-card--with-full-config
> 
> image
> Flip the list icon, or replace it with `favorite` or `bookmark` (which are two more commonly used icons for scorecard)
> 
> I might let the React side slipped... please double check

I'm just going to replace the list icon.   RTL-sensitive icons have no place in these demos since they're too difficult to manage.

emclaug2:
This PR now fixes #202 

- Fixed Drawer shadow, now uses mat-elevation-z4
- Added shadow to UserMenu menu (mat-elevation-z8)
- Removed 4px offset margin to UserMenu header avatar

Available here: https://pxblue-angular-library.web.app

emclaug2:
> https://pxblue-angular-library.web.app/?path=/story/components-info-list-item--with-full-config
> image
> Left icon not center-aligned
> 
> Same for the first info list item of https://pxblue-angular-library.web.app/?path=/story/components-info-list-item--within-a-full-list
> image
> All the other icons of ILI appear fine to me

I don't think this is a component issue, I could add an offset margin to the image to straighten it up.

joebochill:
> I don't think this is a component issue, I could add an offset margin to these problem icons to straighten them.

It definitely looks like a component issue, but related to using `<i>` for icons instead of `<mat-icon>`. I'd be happy if we just swap those out.

joebochill:
Changes look excellent to me 👍 

The only thing I noticed seemed to be a slight inconsistency with the divider props on the drawers being on by default sometimes and off others.

emclaug2:
> Changes look excellent to me 👍
> 
> The only thing I noticed seemed to be a slight inconsistency with the divider props on the drawers being on by default sometimes and off others.

I think nested nav item story is the only instance that has the dividers on by default;  pretty sure this was intentional to match react + better show nav depth.

emclaug2:
> > I don't think this is a component issue, I could add an offset margin to these problem icons to straighten them.
> 
> It definitely looks like a component issue, but related to using `<i>` for icons instead of `<mat-icon>`. I'd be happy if we just swap those out.

Looking closer the icon container is centered, but the icon itself looks shifted up.   I can swap this out with a different mat-icon but it won't match react anymore. 

image


I'll swap this icon out with a mat-icon.

emclaug2:
`with full config` and `within a list` uses devices icon instead of the off-centered icon from before. 

https://pxblue-angular-library.web.app/?path=/story/components-info-list-item--with-full-config

If this is looking good let's get this thing merged in.

huayunh:
This is not fixed or logged
https://github.com/pxblue/angular-component-library/pull/210#issuecomment-766091719

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-info-list-item--within-a-full-list
Use an actual temperature icon
image

emclaug2:
> This is not fixed or logged
> [#210 (comment)](https://github.com/pxblue/angular-component-library/pull/210#issuecomment-766091719)

This is a RTL issue in the storybook addon; RTL dir attribute not being set correctly in this case. 

https://github.com/pxblue/storybook-addons/issues/15

emclaug2:
> https://pxblue-angular-library.web.app/?path=/story/components-info-list-item--within-a-full-list
> Use an actual temperature icon
> image

Temperature icon added: https://pxblue-angular-library.web.app/?path=/story/components-info-list-item--within-a-full-list

huayunh:
dark theme rail footer, try using the eaton logo from our logo library

https://pxblue-angular-library.web.app/?path=/story/components-drawer--with-a-nav-rail

image

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-score-card--with-score-badge

dark theme scorecard, make the badge background to match the card background
image

huayunh:
If drawer rail active item can change color to blue 500, then everything looks good style-wise.

image

huayunh:
Good work Evan, and thanks for the patience :P

emclaug2:
> If drawer rail active item can change color to blue 500, then everything looks good style-wise.
> 
> image

This is working in the real dev storybook, so I'm trying to figure out what changed here to cause the theme missing in the rail.

emclaug2:
> This is working in the real dev storybook, so I'm trying to figure out what changed here to cause the theme missing in the rail.

Okay this took some snooping around; here's what's going on:

PX Blue dev ng storybook is not using latest theme @ v5.2.0.  It's using v5.1.2 because the yarn.lock was unchanged, thus preventing dev from using latest theme.  

v5.2.0 introduces a more-specific style rule for applying active nav item font color: 

`
// v5.2.0
.pxb-drawer-nav-item-active .pxb-info-list-item .mat-list-item-content {
    color: map-get($primary, 500);
}
`
I'm not sure what the motivation here was.

This is a bad change; this only applies to _normal_ nav items.  Rail Nav Items are not inside InfoListItems, so this rule doesn't apply to them.

I'm making a new PR in the themes repo that'll fix this using the css below:

`
.pxb-drawer-nav-item-active, /*NavRail item*/
.pxb-drawer-nav-item-active .pxb-info-list-item .mat-list-item-content /*Normal NavItem*/ {
    color: map-get($primary, 500);
}
`

This'll apply the correct active color to rail nav items.

huayunh (approved):
Change license agreement icon to something that makes more sense
image

Looking good otherwise

PR #221, opened 2021-02-03
Update to latest theme and font packages

joebochill:
- Use latest theme package
- Update font import syntax


emclaug2:
This storybook has the same font issue as React storybook.  Should be the same solution for both.

emclaug2 (approved):

PR #226, opened 2021-02-09
Add default margin to EmptyState action button icon

emclaug2:
- Moved storybook styles into component library as default.
- EmptyState action button icon now has some default margins applied to it. 
- Default action button icon font size is 20px.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 
 -   Default styles of `<pxb-dropdown-toolbar>` for more padding around the dropdown arrow.
 -   Changed `<pxb-user-menu>` menu title default font size to 16px.
+-   Changed default `<pxb-empty-state>` button icon default margin. 
 
 ## v4.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@
 
 -   Default styles of `<pxb-dropdown-toolbar>` for more padding around the dropdown arrow.
 -   Changed `<pxb-user-menu>` menu title default font size to 16px.
--   Changed default `<pxb-empty-state>` button icon default margin. 
+-   Changed default `<pxb-empty-state>` action button icon margins. 
+-   Changed default `<pxb-empty-state>` action button icon size to 20px.
 
 ## v4.0.0
 
--- a/components/src/core/empty-state/empty-state.component.scss
+++ b/components/src/core/empty-state/empty-state.component.scss
@@ -1,3 +1,8 @@
+[dir='rtl'] .pxb-empty-state .pxb-empty-state-actions-wrapper mat-icon {
+    margin-right: -4px;
+    margin-left: 8px;
+}
+
 .pxb-empty-state {
     display: flex;
 }
@@ -36,8 +41,17 @@
         }
     }
 
-    .pxb-empty-state-actions-wrapper .mat-button-wrapper {
-        display: flex;
-        align-items: center;
+    .pxb-empty-state-actions-wrapper {
+        .mat-button-wrapper {
+            display: flex;
+            align-items: center;
+        }
+        mat-icon {
+            margin-right: 8px;
+            margin-left: -4px;
+            height: 20px;
+            width: 20px;
+            font-size: 20px;
+        }
     }
 }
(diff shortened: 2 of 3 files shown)


emclaug2:
image

huayunh:
Looks good visually.

joebochill (commented):
I'm a little confused why we are hacking the styles of user-provided content. Should they not be able to supply whatever content or button styles they want?

And why are we overriding (via negative margins) the default button padding (16px L/R) in this particular case?

huayunh:
Negative margin: angular’s button does not have the concept of a “startIcon” like react does as far as I remember, so a negative margin is necessary to to have everything properly aligned with Material Design’s design spec.

emclaug2:
> I'm a little confused why we are hacking the styles of user-provided content. Should they not be able to supply whatever content or button styles they want?
> 
> And why are we overriding (via negative margins) the default button padding (16px L/R) in this particular case?

The EmptyState style overrides are required to match the design we're showing in our storybook `EmptyState with action` example.   To minimize the amount of assumptions we make about what a user is putting in their button, I've scoped these style overrides to only apply to mat-icons, since they should all come with the same default styles.  

This PR was inspired by this comment: https://github.com/pxblue/angular-showcase-demo/pull/13#discussion_r572879656

joebochill:
> Negative margin: angular’s button does not have the concept of a “startIcon” like react does as far as I remember, so a negative margin is necessary to to have everything properly aligned with Material Design’s design spec.

Okay, fair point. I did not notice before that icon buttons use 12px padding in the material spec. 

But that raises the question, why are these rules in the Empty State component? Should we not be doing this globally in the theme to apply to all mat-button icons, not just ones that are passed to the empty-state component?

huayunh:
I'm more happy to move them to themes and calling it a breaking change.

joebochill:
> I'm more happy to move them to themes and calling it a breaking change.

It's not a breaking change is it? Usage will continue as before, just the styles will be tweaked.

emclaug2:
> > I'm more happy to move them to themes and calling it a breaking change.
> 
> It's not a breaking change is it? Usage will continue as before, just the styles will be tweaked.

This is a non-breaking change.   You're correct.

emclaug2:
Should we not be doing this globally in the theme to apply to all mat-button icons, not just ones that are passed to the empty-state component?

This is something that I would try to avoid.  If you start applying out-of-box styles on angular components at a global scope, you're going to upset someone eventually.  

What do you think about introducing a directive or class name to apply to mat-icon-buttons that'll do these styles for you?  

`pxb-mat-icon` ? 

I know it's  not as convenient, but a theme should be primarily focused on colors and attempt to rearrangement/pad/margin my elements.

joebochill:
> > Should we not be doing this globally in the theme to apply to all mat-button icons, not just ones that are passed to the empty-state component?
> 
> This is something that I would try to avoid. If you start applying out-of-box styles on angular components at a global scope, you're going to upset someone eventually.
> 
> What do you think about introducing a directive or class name to apply to mat-icon-buttons that'll do these styles for you?
> 
> `pxb-mat-icon` ?
> 
> I know it's not as convenient, but a theme should be primarily focused on colors and attempt to rearrangement/pad/margin my elements.

I completely agree...Angular Material should be styling their buttons correctly to match their Material Spec (how is react better at implementing the google spec than google???). So I would leave it out of the theme and out of the empty state entirely (doing it in one place and not in the other is just going to cause confusion and inconsistencies)...if the user wants smaller icons, they can pass in a button with smaller icons. My two cents.

emclaug2:
> I completely agree...Angular Material should be styling their buttons correctly to match their Material Spec (how is react better at implementing the google spec than google???). So I would leave it out of the theme and out of the empty state entirely (doing it in one place and not in the other is just going to cause confusion and inconsistencies)...if the user wants smaller icons, they can pass in a button with smaller icons. My two cents.

Ok I can add the manual styles back into the Showcase & Storybook demos.

huayunh:
MUI is MUCH better than Angular Material 🤦‍♀️  

I must say it is a shame on Sir Google for having a great design system but with an awkward component library

emclaug2:
I'm just going to close this PR since this work is just being reverted.

PR #227, opened 2021-02-10
Update knob text to match API

emclaug2:
- Update knob text to match actual API, even though we are going to change it.


daileytj (approved):

joebochill (commented):
I would just not show a knob for this at all. It's not documented and we're going to be getting rid of it.

joebochill (approved):

PR #231, opened 2021-02-17
Bug/split drawer layout story

### LINKED ISSUE #216: Angular rail detector set to on mouse out
#### Describe the bug

The navigation rail's active item color doesn't change until the user moves the mouse out. This only seems to happen when a rail is inside the drawerLayout.

Should behave.

1. Go to https://pxblue-components.github.io/angular-dev/?path=/story/components-drawer--within-a-drawer-layout 
2. Set variant to rail
3. Click around but don't move the mouse cursor out of the rail area. Observe that the color won't change
4. Move the mouse out. Now you see the color applied correctly.

https://user-images.githubusercontent.com/8997218/105866264-d0b85400-5fc1-11eb-8720-49103860a080.mov

#### Environment
MacOS Chrome
		
#### Anything else to add?
Only seen on storybook so far.

emclaug2:
Fixes #216 

- Doesn't fix the issue, but avoids it. 
- Split `within Drawer layout` story into 2. 

### DIFF
--- a/demos/storybook/stories/drawer/_module.stories.ts
+++ b/demos/storybook/stories/drawer/_module.stories.ts
@@ -20,6 +20,7 @@ import { withMultiNavGroups } from './with-multiple-nav-groups.stories';
 import { withFooter } from './with-footer.stories';
 import { MatDividerModule } from '@angular/material/divider';
 import { withNestedNavItems } from './with-nested-nav-items.stories';
+import { withinDrawerLayoutRail } from './within-drawer-layout-rail.stories';
 
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { withFullConfig } from './with-full-config.stories';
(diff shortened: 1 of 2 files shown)


demos/storybook/stories/drawer/_module.stories.ts
```diff
@@ -113,4 +113,5 @@ storiesOf(`${COMPONENT_SECTION_NAME}/Drawer`, module)
     .addDecorator(storyWrapper())
     .addParameters({ ...STORY_PARAMS, notes: { markdown: getReadMe('Drawer.md') } })
     .add(README_STORY_NAME, getReadMeStory)
-    .add('within a Drawer Layout', withinDrawerLayout);
+    .add('within a Drawer Layout', withinDrawerLayout)
```
joebochill, line 116: Why do we have two different storiesOf calls for the Drawer?

joebochill (approved):

demos/storybook/stories/drawer/_module.stories.ts
emclaug2: All the Drawer stories have a special wrapper (`drawerWrapper`) around it that'll shift canvas content onto the left-side of the screen.   This wrapper doesn't apply to the DrawerLayout stories and will mess up its styles.

PR #234, opened 2021-03-03
Feature/use rems

emclaug2:
- This converts all components' vertical spacing to use rems so that our component library scales nicely with large font-sizes.
- Default should look the exact same. 
- This also updates the Scorecard to use 20px header, 14px subtitle. 

### DIFF
--- a/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
+++ b/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
@@ -1,4 +1,3 @@
-
 /* RTL */
 [dir='rtl'] .pxb-drawer-nav-item-content {
     .pxb-drawer-nav-item-active-highlight {
--- a/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
+++ b/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
@@ -1,3 +1,4 @@
+
 /* RTL */
 [dir='rtl'] .pxb-drawer-nav-item-content {
     .pxb-drawer-nav-item-active-highlight {
@@ -24,7 +25,6 @@ pxb-drawer-nav-item {
 }
 
 .pxb-drawer-condensed-rail .pxb-drawer-nav-item-content .pxb-drawer-nav-item-rail {
-    width: 3.5rem;
     min-height: 3.5rem;
     .pxb-drawer-nav-item-rail-text {
         display: none;
@@ -147,9 +147,6 @@ pxb-drawer-nav-item {
     .pxb-info-list-item .mat-list-item-content {
         padding-left: 16px;
     }
-    /*pxb-info-list-item .pxb-info-list-item .mat-list-text .pxb-info-list-item-title-wrapper {
-        font-weight: 400;
-    }*/
     .mat-expansion-panel-body {
         padding: 0;
     }
--- a/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
+++ b/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
@@ -3,7 +3,7 @@
     .pxb-drawer-nav-item-active-highlight {
         right: 0;
         &.pxb-drawer-nav-item-active-round {
-            border-radius: 24px 0 0 24px;
+            border-radius: 1.5rem 0 0  1.5rem;
         }
     }
     .pxb-drawer-nav-item-depth-3 {
@@ -12,7 +12,7 @@
     }
     pxb-info-list-item.pxb-drawer-nav-item-no-icon-closed .pxb-info-list-item .mat-list-item-content {
         padding-right: 16px;
-        padding-left: 24px;
+        padding-left: 32px;
     }
 }
 
@@ -24,8 +24,8 @@ pxb-drawer-nav-item {
 }
 
 .pxb-drawer-condensed-rail .pxb-drawer-nav-item-content .pxb-drawer-nav-item-rail {
-    width: 56px;
-    min-height: 56px;
+    width: 3.5rem;
+    min-height: 3.5rem;
     .pxb-drawer-nav-item-rail-text {
         display: none;
     }
@@ -37,12 +37,12 @@ pxb-drawer-nav-item {
     user-select: none;
 
     &.pxb-drawer-nav-item-compact {
-        margin-top: 8px;
-        margin-bottom: 8px;
+        margin-top: .5rem;
+        margin-bottom: .5rem;
         .pxb-info-list-item-content {
             &.pxb-info-list-item-dense,
             .mat-list-item-content {
-                height: 40px !important;
+                height: 2.5rem !important;
             }
         }
     }
@@ -91,8 +91,8 @@ pxb-drawer-nav-item {
         left: 0;
         top: 0;
         &.pxb-drawer-nav-item-active-round {
-            width: calc(100% - 8px);
-            border-radius: 0 24px 24px 0;
+            width: calc(100% - .5rem);
+            border-radius: 0 1.5rem 1.5rem 0;
         }
         &.pxb-drawer-nav-item-active-square {
             border-radius: 0;
@@ -115,11 +115,11 @@ pxb-drawer-nav-item {
         z-index: 1;
         display: flex;
         flex-direction: column;
-        min-height: 72px;
-        width: 72px;
+        min-height: 4.5rem;
+        width: 4.5rem;
         justify-content: center;
         align-items: center;
-        padding: 16px 8px;
+        padding: 1rem .5rem;
         box-sizing: border-box;
         cursor: pointer;
         text-align: center;
@@ -127,8 +127,8 @@ pxb-drawer-nav-item {
             z-index: 1; // Keep items above the highlight.
         }
         .pxb-drawer-nav-item-rail-text {
-            font-size: 12px;
-            line-height: 16px;
+            font-size: .75rem;
+            line-height: 1rem;
             word-break: break-word;
             hyphens: auto;
         }
--- a/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
+++ b/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
@@ -12,7 +12,7 @@
     }
     pxb-info-list-item.pxb-drawer-nav-item-no-icon-closed .pxb-info-list-item .mat-list-item-content {
         padding-right: 16px;
-        padding-left: 32px;
+        padding-left: 24px;
     }
 }
 
@@ -123,10 +123,10 @@ pxb-drawer-nav-item {
         display: flex;
         flex-direction: column;
         min-height: 4.5rem;
-        width: 4.5rem;
+        width: calc(3.5rem + 16px);
         justify-content: center;
         align-items: center;
-        padding: 1rem 0.5rem;
+        padding: 1rem 8px;
         box-sizing: border-box;
         cursor: pointer;
         text-align: center;
--- a/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
+++ b/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
@@ -1,3 +1,5 @@
+@import '../../../utility/utility';
+
 /* RTL */
 [dir='rtl'] .pxb-drawer-nav-item-content {
     .pxb-drawer-nav-item-active-highlight {
@@ -101,9 +103,7 @@ pxb-drawer-nav-item {
 
     .pxb-drawer-nav-item-expand-icon {
         transition: 300ms;
-        font-size: 1.5rem;
-        height: 1.5rem;
-        width: 1.5rem;
+        @include scale-icon-content();
         &.pxb-drawer-nav-item-expanded {
             transform: rotate(180deg);
         }
@@ -113,9 +113,7 @@ pxb-drawer-nav-item {
         display: flex;
         flex-direction: column;
         .mat-icon {
-            font-size: 1.5rem;
-            height: 1.5rem;
-            width: 1.5rem;
+            @include scale-icon-content();
         }
     }
     .pxb-drawer-nav-item-rail {
--- a/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
+++ b/components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
@@ -3,7 +3,7 @@
     .pxb-drawer-nav-item-active-highlight {
         right: 0;
         &.pxb-drawer-nav-item-active-round {
-            border-radius: 1.5rem 0 0  1.5rem;
+            border-radius: 1.5rem 0 0 1.5rem;
         }
     }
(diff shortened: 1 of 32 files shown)


components/src/core/drawer/drawer-body/nav-item/drawer-nav-item.component.scss
```diff
@@ -110,25 +112,30 @@ pxb-drawer-nav-item {
...
+        }
     }
     .pxb-drawer-nav-item-rail {
         z-index: 1;
         display: flex;
         flex-direction: column;
-        min-height: 72px;
-        width: 72px;
+        min-height: 4.5rem;
+        width: 4.5rem;
```
emclaug2, line 126: Looks like the Nav Rail (default) has a 4.5rem width.  We adjusted the condensed/closed-persistent rail width to be calc(1.5rem + 16px * 2), so we might have to tweak this a bit.

emclaug2: I've updated the width here to be 3.5rem + 16px.

components/src/core/info-list-item/info-list-item.component.scss
```diff
@@ -187,7 +199,7 @@ mat-divider {
         position: relative;
         bottom: 0;
         right: 0;
-        left: 72px;
-        width: calc(100% - 72px);
+        left: 4.5rem;
+        width: calc(100% - 4.5rem);
```
emclaug2, line 203: This looks funky.

huayunh (commented):
Looks fine visually

https://pxblue-angular-library.web.app/?path=/story/components-drawer--within-a-drawer-layout-as-rail

image

May want to show a regularly styled app bar, and a larger tooltip font size (if that is controlled by us)

And the menu icon is right-shifted a little bit
image
Screenshot from https://pxblue-angular-library.web.app/?path=/story/components-drawer--with-nested-nav-items

emclaug2:
@huayunh That drawer menu icon is giving some trouble, I'm trying to figure out how to center it still; might require some math.  

I'll update that `within a drawerlayout with nav rail` example to use a skinnier toolbar.

emclaug2:
> I'll update that `within a drawerlayout with nav rail` example to use a skinnier toolbar.

I can't make the toolbar skinnier without changing the title font-size, which I think is fine.

emclaug2:
@huayunh The mat-tooltips actually don't respond to in-app system font changes.  You have to set the system font size then refresh the page to see the new tooltip size.

huayunh:
> I'll update that within a drawerlayout with nav rail example to use a skinnier toolbar.

Yeah it is a tricky one.

You will need to make the `padding-left` of 
image
to be always `16px`, and the `margin-left` of 
image
to be always `0.5rem`

And accommodate for RTL directions when you tweak it this way

huayunh:
> I can't make the toolbar skinnier without changing the title font-size, which I think is fine.

This is what it looks like right now in regular size: 
image

emclaug2:
Thanks for the fix @huayunh ! The drawer menu icon should be center aligned now.

emclaug2:
> This is what it looks like right now in regular size: 

I can't identify the problem.  Here's the extra large size:

image

image

huayunh:
If you see how large the size is in the current ng storybook... 
https://pxblue-components.github.io/angular-dev/?path=/story/components-drawer--within-a-drawer-layout-as-rail

image

emclaug2:
So make the font-size smaller in the app-bar?

demos/storybook/stories/drawer/within-drawer-layout-rail.stories.ts
```diff
@@ -34,17 +43,16 @@ export const withinDrawerLayoutRail = (): any => ({
...
-                    style="padding: 0 24px">
+                <mat-toolbar [style.backgroundColor]="blue" [style.color]="white" class="app-bar">
                     <button *ngIf="variant === 'temporary'" mat-icon-button 
                         [style.marginRight.px]="direction() === 'rtl' ? -16 : 16"
                         [style.marginLeft.px]="direction() === 'rtl' ? 16 : -16"
                         (click)="state.open = true">
                         <mat-icon>menu</mat-icon>
                     </button>
-                    <h2>Drawer Layout Demo</h2>
+                    <h2 class="mat-display-1">Drawer Layout Demo</h2>
```
huayunh, line 53: I see what's happening. This should not be `mat-display-1`... 

React's H6 should correspond to `.mat-h2` or `.mat-title` in ng Material

emclaug2: You got it! It'll be deployed in a minute.

huayunh:
## Dropdown toolbar
image
should scale by rem

## Scorecard
image
This padding should be `1rem 16px 0`, not `1rem 1rem 0`

## User Menu
I guess the size of the right avatar is not controlled by us?
image

## User Menu 2
Same trouble with the toolbar text in 
https://pxblue-angular-library.web.app/?path=/story/components-user-menu--within-a-toolbar

emclaug2:
I noticed the header text was not aligning properly in the usermenu, so I had to update some of the margins to use a calc.

emclaug2:
Good eyes @huayunh , all changes will be deployed shortly.

emclaug2:
I updated the UserMenu's styles so that the icon avatar passed in will scale with system font.

demos/storybook/stories/user-menu/within-a-toolbar.stories.ts
```diff
@@ -5,14 +5,17 @@ import { invertRTL } from '../../src/utils';
 export const withinToolbar = (): any => ({
     template: `
         <div>
-            <mat-toolbar style="padding: 0px 16px">
-                <h2>Toolbar Title</h2>
+            <ng-template #avatar>
+                    <mat-icon>person</mat-icon>
+            </ng-template>
+            <mat-toolbar style="padding: 0 16px; height: 4rem;">
+                <h2 class="mat-display-1">Toolbar Title</h2>
```
huayunh, line 12: You gotta change this too.

huayunh (commented):
# 1
If you change the system font size to extra large, then in https://pxblue-angular-library.firebaseapp.com/?path=/story/components-drawer--within-a-drawer-layout, change the variant to temporary, and you will see a tiny little menu icon. Should use rem.

# 2
On an "extra small" setting, if you go to https://pxblue-angular-library.firebaseapp.com/?path=/story/components-list-item-tag--within-an-infolistitem, you will see that the tags are overflowing... I might prefer to use a margin instead of flex box space-between in this case
image

---
Everything else looks good to me visually. I checked both on extra small, extra large, and RTL.

demos/storybook/stories/user-menu/within-a-toolbar.stories.ts
emclaug2: Updated now :)

emclaug2:
Thanks for the detailed review @huayunh , all changes are now deployed: 
https://pxblue-angular-library.web.app

huayunh:
The # 1 mentioned appears unchanged... the tiny menu icon is still there.

image
(Screenshot from 2x)

For # 2, you may just do `margin: 0 16px` for this to accommodate the RTL, or they will get stacked together
image

huayunh:
And remember to `git push` after the deploy

emclaug2:
Ooof, forgot to fix the little menu icon.  I'm on it.

components/src/core/drawer/drawer-layout/drawer-layout.component.scss
```diff
@@ -9,7 +9,7 @@ $transition: 300ms cubic-bezier(0.4, 0, 0.2, 1);
 [dir='rtl'] .pxb-drawer-layout-content .mat-drawer-side {
     border-left: 0; // Border-left is provided by the Drawer component.
     .pxb-drawer-layout-sidenav.pxb-drawer-layout-shadow {
-        box-shadow: 0 0 -4px 0 #424e54;
+        box-shadow: 0 0 -0.25rem 0 #424e54;
```
huayunh, line 12: I don't think we will need this for the box-shadow... Would be better if this logic is moved to the component itself, and do something like `[class.mat-elevation-z6]="!sideBorder"`

components/src/core/drawer/drawer-layout/drawer-layout.component.scss
```diff
@@ -37,7 +37,7 @@ $transition: 300ms cubic-bezier(0.4, 0, 0.2, 1);
     }
 
     .pxb-drawer-layout-sidenav.pxb-drawer-layout-shadow {
-        box-shadow: 0 0 4px 0 #424e54;
+        box-shadow: 0 0 0.25rem 0 #424e54;
```
huayunh, line 40: Same for this guy

components/src/core/hero/hero.component.scss
```diff
@@ -12,17 +12,17 @@
...
     display: flex;
-    flex: 1 1 0px;
+    flex: 1 1 0;
     flex-direction: column;
     align-items: center;
     justify-content: flex-start;
     font-size: 1rem;
     overflow: hidden;
-    padding: 16px 8px;
+    padding: 1rem 0.5rem;
```
huayunh, line 21: why 0.5rem for horizontal padding instead of 8px?

components/src/core/score-card/score-card.component.scss
```diff
@@ -38,34 +33,37 @@
...
     .pxb-score-card-action-items-wrapper {
         text-overflow: unset;
         overflow: unset;
-        margin-right: -12px;
+        margin-right: -0.75rem;
         > * {
             cursor: pointer;
             margin-right: 12px;
             text-align: center;
+            @include scale-icon-content();
```
huayunh, line 48: For the very right icon, the right margin should be 16px.
image

components/src/core/score-card/score-card.component.scss
```diff
@@ -76,8 +74,8 @@
     align-items: center;
     .pxb-score-card-badge-wrapper {
         z-index: 1;
-        margin-left: 16px;
-        margin-right: 16px;
+        margin-left: 1rem;
+        margin-right: 1rem;
```
huayunh, line 78: horizontal margin should be fixed value, 16px, to save some more space here

demos/storybook/stories/user-menu/within-a-toolbar.stories.ts
```diff
@@ -5,14 +5,17 @@ import { invertRTL } from '../../src/utils';
...
                 <pxb-spacer></pxb-spacer>
                 <pxb-user-menu menuTitle="Jane Doe" menuSubtitle="Account Manager" [(open)]="open">
-                    <mat-icon pxb-avatar>person</mat-icon>
-                    <mat-icon pxb-menu-avatar>person</mat-icon>
+                    <template pxb-avatar [ngTemplateOutlet]="avatar"></template>
+                    <template pxb-menu-avatar [ngTemplateOutlet]="avatar"></template>
                     <div pxb-menu-body>
```
huayunh, line 17: The `<div style="font-size: 16px; padding: 16px; height: 100px;">Body Content Goes Here</div>` line will appreciate a `mat-body-1` class

components/src/core/drawer/drawer-layout/drawer-layout.component.scss
huayunh: the purpose of a box shadow is to add elevation to a surface. I don't think people needing extra large font size also need an "extra deep" UI

emclaug2: This is a good idea; I think this was done on the Drawer component but not on the Drawer Layout.   Good catch.

components/src/core/hero/hero.component.scss
emclaug2: No reason, this is updated now.

components/src/core/score-card/score-card.component.scss
emclaug2: Updated this.

components/src/core/score-card/score-card.component.scss
emclaug2: Updated this,

demos/storybook/stories/user-menu/within-a-toolbar.stories.ts
emclaug2: You got it.

emclaug2:
The drawer box-shadow has been replaced with the mat-elevation-z6 class.  The scorecard component action-items ng-content is extremely finiky, I'm making an issue to make it a little more robust.

emclaug2:
> The scorecard component action-items ng-content is extremely finiky, I'm making an issue to make it a little more robust.

https://github.com/pxblue/angular-component-library/issues/235

emclaug2:
All changes are deployed (and pushed) now :)
https://pxblue-angular-library.web.app/?path=/story/*

emclaug2:
image

huayunh:
https://pxblue-angular-library.web.app/?path=/story/components-drawer--with-a-nav-rail 
This one is missing a shadow (or a border)

emclaug2:
> https://pxblue-angular-library.web.app/?path=/story/components-drawer--with-a-nav-rail
> This one is missing a shadow (or a border)

But this one never had one to begin with; if it's in the DrawerLayout then it will have a shadow or border.

huayunh:
Hmm... that sounds like a strange API design. I would expect that directive to be on all variants of drawers.

demos/storybook/stories/list-item-tag/within-an-InfoListItem.stories.ts
```diff
@@ -6,9 +6,18 @@ export const withinAnInfoListItem = (): any => ({
...
-            <div pxb-right-content style="width: 160px; display: flex; justify-content: space-between;">
-                <pxb-list-item-tag label="Build Passing" [backgroundColor]="colors.green[300]" [fontColor]="colors.black[900]"></pxb-list-item-tag>
-                <pxb-list-item-tag label="5 Bugs" [backgroundColor]="colors.red[300]" [fontColor]="colors.black[900]"></pxb-list-item-tag>
+            <div pxb-right-content>
+                <pxb-list-item-tag 
+                    label="Build Passing" 
+                    [style.marginRight.px]="16"
+                    [backgroundColor]="colors.green[300]" 
+                    [fontColor]="colors.black[900]">
+                </pxb-list-item-tag>
```
huayunh, line 15: set marginLeft to 16 too for RTL

emclaug2: Now I remember why I did this as a flex space-inbetween.

huayunh:
Everything else looks good to me

emclaug2:
> Hmm... that sounds like a strange API design. I would expect that directive to be on all variants of drawers.

All stand-alone Drawer stories feature the Drawer component outside of the DrawerLayout; the `with a rail` story actually wraps the Drawer in a DrawerLayout so we can apply the rail variant.  The lack of shadow is being caused by lack of side-nav content.  If you open the `within a drawerlayout +rail` story then the shadow works as expected.

demos/storybook/stories/list-item-tag/within-an-InfoListItem.stories.ts
emclaug2: Ok this is fixed now, deploy incoming.

huayunh (approved):
I will log that as a separate issue

PR #237, opened 2021-03-04
Add a sideborder

### LINKED ISSUE #236: Rail's sideBorder / shadow should be on the component itself

When `sideBorder` is set to off, the drawer nav rail should have a shadow; otherwise a side border should be displayed

If you go to https://pxblue-components.github.io/angular-dev/?path=/story/components-drawer--with-a-nav-rail, you will notice that the component has neither a side border nor a shadow.

But this story is OK: https://pxblue-components.github.io/angular-dev/?path=/story/components-drawer--within-a-drawer-layout-as-rail

It came from this PR: https://github.com/pxblue/angular-component-library/pull/234/files#issuecomment-790652706

> All stand-alone Drawer stories feature the Drawer component outside of the DrawerLayout; the `with a rail` story actually wraps the Drawer in a DrawerLayout so we can apply the rail variant.  The lack of shadow is being caused by lack of side-nav content.  If you open the `within a drawerlayout +rail` story then the shadow works as expected.

emclaug2:
Fixes #236 

- Add sideborder to rail story.

A drawer rail cannot exist outside of a DrawerLayout because the `variant` prop is on the DrawerLayout.  The shadow prop is not working in this example because there's no content body to cast the shadow onto.  I've just added a `sideBorder=true` to add a border. 

image

### DIFF
--- a/demos/storybook/stories/drawer/with-nav-rail.stories.ts
+++ b/demos/storybook/stories/drawer/with-nav-rail.stories.ts
@@ -56,6 +56,6 @@ export const withNavRail = (): any => ({
     props: {
         navItems: navItems,
         condensed: boolean('condensed', true),
-        divider: boolean('divider', false)
+        divider: boolean('divider', false),
     },
 });
--- a/demos/storybook/stories/drawer/with-nav-rail.stories.ts
+++ b/demos/storybook/stories/drawer/with-nav-rail.stories.ts
@@ -33,7 +33,7 @@ export const withNavRail = (): any => ({
     ],
     template: `
         <pxb-drawer-layout variant="rail">
-            <pxb-drawer pxb-drawer [condensed]="condensed">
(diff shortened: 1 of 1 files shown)


joebochill (commented):
I don't know if this addresses the linked issue. If the showBorder is false or not specified, we should see a shadow.

emclaug2:
> I don't know if this addresses the linked issue. If the showBorder is false or not specified, we should see a shadow. Why would the shadow visibility be affected by the absence or presence of other page content?

Storybook be default squishes the canvas to take up as little space as possible; in order to show the shadow, we need to add some additional padding to the right of the rail so the shadow can spread.   I can add this padding, just seems a little hacky.  Outside the context of storybook the border would work.... assuming there's right-content.

joebochill (approved):

PR #238, opened 2021-03-09
Feature/dropdown toolbar color

### LINKED ISSUE #233: DropdownToolbar Variant Prop
Add the ability to change the dropdown toolbar `mat-tool` color from primary|warn|error|undefined

There's no way for me to update this. 

No

emclaug2:
Fixes #233 

- Add a `color` prop to the DropdownToolbar

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 -   Adds new property `color` to `<pxb-dropdown-toolbar>`.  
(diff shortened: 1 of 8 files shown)


daileytj (approved):

joebochill (approved):
Are there any other components where we might want to include this while we are in here? Anything else that uses a top-level angular material component that supports multiple variants?

PR #245, opened 2021-03-30
[NG-#1]Update package version number to 4.2.0

daileytj:
Update package version number to 4.2.0

*Note: prettier reformatted this changelog differently than the other repo's for some reason unknown to me. Is this intended or should I manually update this to have the same format as the other changelogs?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,138 +4,138 @@
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


huayunh (approved):
Eyeballing code looks good

joebochill (approved):

joebochill:
prettier looks like it was set up in the components folder, but not at the root level...hence why it missed the changelog.

Issue #246, opened 2021-03-30
Upgrade to latest version of Angular


daileytj:
#### Describe the desired behavior
Use the latest version of Angular

#### Describe the current behavior
Not using the latest version of Angular

huayunh:
Closing this; we have a release branch

PR #257, opened 2021-04-13
Feature/upgrade angular 11

daileytj:
Fixes PXBLUE-2027 & #246 .

- Upgrade components to Angular 11


emclaug2 (approved):

jeffvg (approved):

jeffvg:
Just update Compatibility table

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@
 
 -   Added new property `disabled` to `<pxb-info-list-item`>.
 
+### Changed
+
+-   Migrated to Angular 11
```
joebochill, line 11:
```suggestion
-   Updated to Angular 11 for building the library
```

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@
...
+
+    _Breaking Changes_
+
+-   Dropped support for IE9, IE10, and IE mobile. This was announced in the Angular [v10 Update](https://blog.angular.io/version-10-of-angular-now-available-78960babd41#c357).
+
```
joebochill, line 16: This style of section is not part of the spec for Keep a Changelog (https://keepachangelog.com/en/1.0.0/). 

Couple questions on this:
- Is this really a breaking change? If so then this should be a major version change, not a patch.
- If this was a change in Angular 10, then why do we mention it here if we have been using Angular 10 to build for months now? Should we move this to the version where we started using Angular 10 to build?

joebochill (changes requested):

CHANGELOG.md
daileytj: So this may not actually be a breaking change. I'm not sure what all they mean by dropping support for ie9, 10, and mobile as they don't give much detail. All they say is "Based on heavy consultation with the community, we are deprecating support for older browsers including IE 9, 10, and Internet Explorer Mobile."

Regarding why we didn't mention it sooner, Angular announced that they would deprecate browser support in the future with their v10 release but it wasn't actually removed until v11 (there's nothing in the documentation stating this though https://angular.io/guide/deprecations#deprecated-apis-and-features, I just seen it via the angular upgrade tool). 
Screen Shot 2021-04-14 at 8 04 37 AM

emclaug2: Good catch @joebochill , this should be a major change (v5.0.0) not a patch since our base framework had a major change.

joebochill: We have never supported IE9 or IE10 (not sure about mobile, but probably not that either). We support IE11, but that's also going to be dropped since it's no longer supported by microsoft as of Nov last year. 

So I don't actually think this is a breaking change. Technically, it should only be a major bump if it breaks something in existing usage or dependencies change. If we can still use these components in an Angular 9 or Angular 10 project, then it's not breaking.

However, I'm okay with setting a convention of bumping the major to have better synchronization with the Angular version and making our compatibility tables a little simpler. If we're going to call this a 5.0.0, then it should be going into a release branch instead of dev. (How's the release planning story coming @emclaug2 😛 )

joebochill:
So I think we are good here, we just need to create a new branch off of dev for `release/v5.0.0` and make this a PR into that branch instead of directly into `dev`.

emclaug2:
Or you could just rename this branch to `release.v5.0.0` since it's already up to date with dev.

daileytj:
Closing this PR in favor of branch release/v5.0.0.

PR #260, opened 2021-04-23
Add publish script

emclaug2:
- Add publish script

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -236,3 +236,4 @@ workflows:
                     branches:
                         only:
                             - master
+                            - dev
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -207,7 +207,7 @@ jobs:
         - run:
             name: Publish @pxblue/angular-components
             command:  |
-              cd dist && yarn publish:package -b $CIRCLE_BRANCH
+              yarn publish:package -b $CIRCLE_BRANCH
 
 workflows:
     version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,19 +201,13 @@ jobs:
             at: .
         - run:
             name: Authenticate with registry
-            command: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/.npmrc
+            command: |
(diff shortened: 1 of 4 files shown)


joebochill (commented):
Code looks good...just see my comment about reorganizing the publishing.md file.

PUBLISHING.md
```diff
@@ -1,13 +1,30 @@
...
+  In both cases, the code will only be published if the version number differs from the current version published under the respective dist tag.
 
+## Manually Publishing
+
+If you need to publish a package manually, you can run the following commands from the root folder:
+
+```sh
+yarn build 
+cd dist
+yarn publish:package
```
joebochill, line 17: Is this script accessible when you are in the dist folder?

joebochill: It might be preferable to have the 'cd' commands as part of the publish:package script itself

emclaug2: Great catch! This isn't a great situation because the ng-packgr build cmd will strip out the scripts section from the package.json.  I'll have to figure out the best way to fix this.

emclaug2: I think it makes more sense to move the publish command out from the components folder, make it top level and move the cd command to the publish cmd like you mentioned.

.circleci/config.yml
```diff
@@ -201,19 +201,13 @@ jobs:
...
-              cd dist
-              MASTER_VERSION=`node -p "require('./package.json').version"`
-              NPM_LATEST_VERSION=`npm show @pxblue/angular-components version`
-              if ! [ $MASTER_VERSION == $NPM_LATEST_VERSION ];
-              then
-                  npm publish
-              else
-                  echo "Latest version is already published."
-              fi
+              cd dist && yarn publish:package -b $CIRCLE_BRANCH
```
joebochill, line 210: I think we can remove this 'cd' here now

emclaug2: Nice catch, forgot about this one.

emclaug2: This is fixed now.

joebochill (approved):

PR #261, opened 2021-04-27
Fix info list item icon alignment

### LINKED ISSUE #252: Info List Item with PX Blue icon appears to be off-centered
PX Blue icons in info list items should be properly vertically centered.

1. Go to https://pxblue-components.github.io/angular-dev/?path=/story/components-info-list-item--within-a-full-list

image

#### Environment
MacOS Chrome
		
#### Anything else to add?
I think we tried attacking this issue before, and there were some problems with the use of icon fonts. Feel free to close this issue if we've tried fixing it and found this to be an external dependency issue.

emclaug2:
Fixes #252 

- Fix info list item icon mis-alignment; it's now centered

Hosted here: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 -   Added new property `disabled` to `<pxb-info-list-item`>.
 
+### Fixed
+
+-   Fixed non-centered icon alignment in `<pxb-info-list-item>`.
+
 ## v4.2.0 (March 31, 2021)
 
 ### Added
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
```
joebochill, line 6: So in the planning we have this listed as a 4.3.0 instead of 4.2.1. If you want to update that and put today's date on it, we should be able to get this published today.

joebochill (approved):

CHANGELOG.md
emclaug2: I've just updated the changelog and package version, should be good to get this out the door today.

joebochill (approved):

PR #270, opened 2021-05-14
Add new prop to disable tooltip

### LINKED ISSUE #268: Ability to turn off tooltips on condensed Rail
I want to be able to enable/disable the tooltips that are applied to the condensed Rail Items.

Currently they are on by default when the rail is condensed and cannot be turned off.

emclaug2:
Fixes #268 

- Add prop to enable the default tooltip hover for rail

What is the naming convention we want to do here for prop names?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v4.4.0 (not published yet)
 
+### Added
+
+-   Added `disableRailTooltip` prop to `<pxb-drawer>`.
+
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


jeffvg (approved):
guess ShowTooltipOnRailHover is good or shorter ShowTooltipOnRail

joebochill (commented):
I agree with Jeff the name is kinda cumbersome. 

The default is enabled, so shouldn't we flip this to be something like `disableRailTooltip`?

Is the tooltip showing for all rail items, or only if we are in the condensed version?

joebochill (approved):

PR #271, opened 2021-05-17
Fix list item tag offcentered text

### LINKED ISSUE #232: ListItemTag Padding off by 1px
#### Describe the bug
List Item Tag
 due to the way letter-spacing works, there is a 1px extra padding on the right. So Instead of doing padding: 0 4px, it should be 0 3px 0 4px, and perhaps get flipped when written RTL.

Padding should look even. 

https://pxblue-components.github.io/angular-dev/?path=/story/components-list-item-tag--within-an-infolistitem

#### Environment

Chrome, Windows

emclaug2:
Fixes #232 

- Fix off-centered list item tag text

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## v4.3.1 (not published yet)
+
+### Fixed
+
+-   Fixed non-centered `<pxb-list-item-tag>` text.
+
 ## v4.3.0 (April 27, 2021)
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## v4.3.1 (not published yet)
+## v4.4.0 (not published yet)
 
 ### Fixed
 
--- a/components/src/core/list-item-tag/list-item-tag.component.scss
+++ b/components/src/core/list-item-tag/list-item-tag.component.scss
@@ -15,6 +15,7 @@
         font-weight: 700;
         text-transform: uppercase;
         letter-spacing: 1px;
+        margin-right: -1px;
         line-height: 1rem;
         font-size: 0.625rem;
     }
--- a/components/package.json
+++ b/components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pxblue/angular-components",
-  "version": "4.3.1",
+  "version": "4.4.0",
   "description": "Angular components for PX Blue applications",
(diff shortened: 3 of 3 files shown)


components/src/core/list-item-tag/list-item-tag.component.scss
```diff
@@ -15,6 +15,7 @@
         font-weight: 700;
         text-transform: uppercase;
         letter-spacing: 1px;
+        margin-right: -1px;
```
joebochill, line 18: should letter spacing be specified in rems to look correct when the font is anything other than 16px?

Does the margin-right approach work when you flip to RTL with a true RTL language?

emclaug2: I checked this with RTL language but the text still had the letter-spacing applied to the right side instead of the left, so it seems this margin adjustment works without RTL special attention.  

For larger font sizes, I have to check that out and see.

emclaug2: Since the letter spacing is set to be 1px, the negative 1px adjustment works with all font sizes.

joebochill: But does the letter spacing itself look correct when we specify 1px? I.e., 1px between letters that are 16px tall looks a certain way, 1px between letters that are 40px tall might not look how we intend.

huayunh: If someone set their system font size to be x2.5 of our default, I would imagine a lot of designs to go pretty off the rail.

`letter-spacing` touches the horizontal spacing, which shall generally stay constant no matter what the default font size is

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## v4.3.1 (not published yet)
```
joebochill, line 3: Shall we roll this up into the upcoming 4.4.0 or are we planning to publish a patch before then?
image

joebochill (approved):

CHANGELOG.md
emclaug2: I think it might be best to adjust the roadmap so 4.4.0 is mobile-friendly use menu, 4.5.0 is collapsible appbar.  The latter will probably take a while.  I have no issues making 4.4.0 include this bug fix, the drawer tooltip hover prop, and the user-menu mobility enhancements.

PR #273, opened 2021-05-25
Feature/mobile user menu

emclaug2:
-  Use BottomSheet on as the overlay when UserMenu is opened on mobile viewports.
- Update docs, storybook, styles for bottomsheet support

The DrawerHeader in the bottomsheet is not the correct color; this color is provided by the @pxblue/angular-themes, so it needs to be updated there instead. 

Hosted here: 
https://pxblue-angular-library.web.app/?path=/story/components-user-menu--with-full-config


emclaug2:
This was a little trickier than expected because the content that appears in the menu is treated as a `<ng-template>` and these templates can only be rendered in 1 place at a time.  So this means the bottomsheet cannot appear until the menu dismissed and vise versa.

huayunh:
There is an undesired round corner 
image

huayunh:
Love to learn how everyone else feels about the different entering/exiting speed of the bottom sheet. Otherwise looks visually good

daileytj:
> Love to learn how everyone else feels about the different entering/exiting speed of the bottom sheet. Otherwise looks visually good

I think it would be nice if the dismiss animation was sped up a little to match the entering speed but I'm not going to call it a deal breaker for merging this.

daileytj:
Screen Shot 2021-05-26 at 9 05 28 AM

@emclaug2 I know you mentioned the header color issue, but there also seems to be a text color issue in the header on the showcase. It doesn't look like any of your changes would effect this and I'm not sure if this is a UserMenu, DrawerHeader, or toolbar themes issue but we might want to log this as a separate issue if it's not a result of a change you made in this PR.

components/src/core/user-menu/user-menu.component.scss
```diff
@@ -39,22 +39,45 @@ mat-card.pxb-user-menu-overlay {
...
+}
 
-.pxb-user-menu-overlay-backdrop {
-    background-color: unset;
+.cdk-overlay-backdrop-showing {
+    &.pxb-user-menu-overlay-backdrop {
+        background-color: unset;
+    }
+    &.pxb-user-menu-bottomsheet-backdrop {
+        background-color: #424e54;
```
daileytj, line 61: Should we be importing the colors package and using gray 900 here instead of the hardcoded value? Do we have a common practice/guideline for this?

daileytj (approved):

emclaug2:
@daileytj  The UserMenu default menu drawerheader shouldn't be affected by these style changes; I'll take a look at the showcase to see why this is happening before merging this in.

components/src/core/user-menu/user-menu.component.scss
emclaug2: Right now our component library doesn't have a dependency on the colors package and this hex code will be overwritten by our theme.  I just picked a reasonable default as a placeholder.  The backdrop above is a good example - without this default-override or our theme-overrides (TODO) the backdrop color is `unset` so I just looked up the hex code for black[500].  

Since our practice so far has been to rely on our themes package to provide color overrides; I don't think this inline-hex code is an issue since it'll be overwritten anyways.  

On the otherhand, if we add the @pxblue/colors package, I think we would start to more commonly set our default colors in the component library instead of the themes.  This would could cause an issue down the road for teams using non-blue themed palettes.   We _could_ add the colors package, but we'd then we'd need to be extra attentive to make sure we don't put any hard-coded values that will go unnoticed in a blue/bluedark theme, but be look out-of-place in a non-blue theme. 

If we could access the applied-theme inside our component library repo, I'd suggest moving our component styles into the comp library repo, but unfortunately we can't do that.    For now I'd say we just add the placeholder styles so non-themed components looks at least decent.

emclaug2:
> @daileytj The UserMenu default menu drawerheader shouldn't be affected by these style changes; I'll take a look at the showcase to see why this is happening before merging this in.

I didn't see the showcase showing any incorrect header colors so I'm going to merge this is and run everything again from dev branch.  If there's an issue I'll just make a small bug fix in the showcase demo instead.

PR #282, opened 2021-06-01
Feature/host feature

emclaug2:
- Hosting POC using firebase 
- Firebase console: https://console.firebase.google.com/u/3/project/pxb-storybook-ng-feature/overview
- Hosted here: https://pxb-storybook-ng-feature.web.app/index.html?path=/story/intro-overview--px-blue-angular-components

Normally the URL ends with `.web.app` but a weird error is happening that requires a `/index.html` at the end; I'll check tomorrow to see if the error persists.


emclaug2:
Here's what I did: 

- I went into develop.pxblue@gmail.com's firebase console.
- Initialized a new project
- Generated an access token
- Put token into circleci env vars
- Deploy from CI/CD

jeffvg (approved):
everything seems good to me.

demos/storybook/package.json
```diff
@@ -10,7 +10,8 @@
         "lint": "eslint \"stories/**/**.ts\"",
         "lint:fix": "eslint \"stories/**/**.ts\" --fix",
         "prettier": "prettier \"**/**.{ts,json,css,scss,html}\" --write",
-        "prettier:check": "prettier \"**/**.{ts,json,css,scss,html}\" --check"
+        "prettier:check": "prettier \"**/**.{ts,json,css,scss,html}\" --check",
+        "deploy:feature": "yarn add firebase-tools && ./node_modules/.bin/firebase deploy --non-interactive --token $FIREBASE_FEATURE_TOKEN"
```
joebochill, line 14: Do we need to have yarn add do a fresh install every time in the command, or should we add firebase-tools to our devDependencies so it can be cached with the rest of our dependencies? Or use npx?

.circleci/config.yml
```diff
@@ -173,8 +173,11 @@ jobs:
 
                       if [ $CIRCLE_BRANCH == $MASTER_BRANCH ]; then
                         git clone $MASTER_TARGET out
-                      else
+                      elif [ $CIRCLE_BRANCH == $DEV_BRANCH ]; then
                         git clone $DEV_TARGET out
+                      else
+                        yarn run deploy:feature
```
joebochill, line 179: Can we add an echo statement in here to say something like:
"Deploying feature branch to firebase: (url)"

emclaug2: Yep you got it!

demos/storybook/package.json
emclaug2: Yes we can add the firebase tools to our devDependencies list so that we can cache this dependency.

emclaug2:
For some unknown reason, the cache is not being restored correctly during the deploy step.  Since there's only 1 dependency required & this dependency is only added for feature-branch workflows, I think it makes sense to leave the deploy command as `yarn add firebase-tools && [publish]`.

joebochill (approved):

PR #284, opened 2021-06-03
Feature/app bar with transition

emclaug2:
- App Bar that transitions from opened to closed when a scroll distance is hit. 
- Storybook demo: https://pxblue-angular-library.web.app/


emclaug2:
TODO: Support a background image & Handle collapsed toolbar height on screen resizing.

huayunh:
I was just thinking that maybe we can make the _default_ scroll distance be like this:

When scrolling down, the snap happens at `y=expandedHeight-collapsedHeight`

When scrolling up, the snap happens at `y=0`

I'm hoping that this can effectively conceal the white margins, and does not annoy the users too much when the app bar suddenly expands and covers up their page content

huayunh:
And the storybook you hosted on Firebase... if you click "app" and then one of the stories, the page content part would stay on the "notes" tab (i.e., readme), instead of automatically switch to the "canvas" tab.
Screen Shot 2021-06-07 at 11 25 32 AM

emclaug2:
This is now ready for another round of review and is hosted here:  https://pxblue-angular-library.web.app


The only thing I forgot to do was adjust the 3-line default positing to be 2/3 of the toolbar height.   I'm working on that now.

emclaug2:
Inner content positioning is now 3/5 of the expanded size for the default font-sizes.  If a user needs to adjust their default top positioning, there is a css class available for them to adjust the defaults.

Originally I wanted the component to be smart enough to calculate the default positioning, but this becomes increasingly difficult whenever the content grows and shrinks.   I made it so the default positioning looks good, but is adjustable.

jeffvg:
I noticed if you adjust the expanded height to less than what we should use in a design that the menu bleeds into the content.
This is collapsedHeight 64 and expandedHeight 125

image

joebochill:
The toolbar itself feels good. The three line component feels a little goofy as Jeff has mentioned. Seems like it requires a lot of user-managed styles to make it work right out of the box.Maybe we can chat about this again today and see how we want to handle this.

emclaug2:
> The toolbar itself feels good. The three line component feels a little goofy as Jeff has mentioned. Seems like it requires a lot of user-managed styles to make it work right out of the box.Maybe we can chat about this again today and see how we want to handle this.

The issue that @jeffvg is facing is that the current toolbar component is not smart enough to follow the design specs & place the center of the content at 3/5 the height of the toolbar component; right now the placement works with the defaults, but a user would have to manually go in and specify the y-offset of the toolbar content in the expanded state.

emclaug2:
I've merged dev into this branch so we now have the automated storybook feature branch deployments.
The latest changes are hosted here:  https://pxb-storybook-ng-feature.web.app

There is now a ThreeLinerComponent in addition to the AppBarComponent.

I had some trouble making the storybook stories for the ThreeLinerComponent interesting because the `with full config` story is essentially the same as the `with basic usage`.  Also the `within context` story requires a little bit of thinking because I'd need to port over the AppBar's story wrapper into a single story.  This is still a todo, but the AppBar should be finalized now.

huayunh:
I assume you would polish the dropdown menu in a future PR?

In this screenshot, the horizontal padding of your app bar (as denoted by the red stroke) should be 16px
Screen Shot 2021-06-15 at 7 50 12 PM

emclaug2:
> I assume you would polish the dropdown menu in a future PR?
> 
> In this screenshot, the horizontal padding of your app bar (as denoted by the red stroke) should be 16px
> Screen Shot 2021-06-15 at 7 50 12 PM

For now I would omit any menu changes; I think as a future story we talked about making that menu as another component.  I'll update the padding to be 16px for the basic storybook examples.

jeffvg:
Just noticing this, should mat-menu-context "text" be right-aligned when in RTL

image

components/src/core/three-liner/three-liner.component.scss
```diff
@@ -0,0 +1,19 @@
...
+.pxb-three-liner-title {
+    font-size: 30px;
+    transition: $transition;
+}
+.pxb-three-liner-subtitle {
+    font-size: 16px;
+    transition: $transition;
+}
+.pxb-three-liner-info {
+    font-size: 14px;
```
joebochill, line 16: These font sizes should be in rem to respond to system font size

docs/ThreeLiner.md
```diff
@@ -0,0 +1,33 @@
+# Three Liner
+
+The `<pxb-three-liner>` accepts content that can grow and shrink within the context of a `<pxb-app-bar>`.
```
joebochill, line 3:
```suggestion
The `<pxb-three-liner>` can display up to three lines of stylized text or other ng-content. It is most commonly used within the context of a `<pxb-app-bar>` where the text can grow / shrink as the App Bar is expanded and collapsed.
```

joebochill: Can you add a link to the app bar component docs as well?

joebochill (commented):
Did you run through some final checks to make sure everything looks good and works as expected in a showcase-style application?

Storybook looks solid.

components/src/core/app-bar/app-bar.component.scss
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.pxb-app-bar-collapsed {
+    .pxb-app-bar-body-wrapper > * {
+        top: 0;
+    }
+}
+.pxb-app-bar-body-wrapper > * {
+    position: relative;
+    top: 80px;
```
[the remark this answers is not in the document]
emclaug2: Yes, you're right - I've  removed the default top 80 and adjusted the non-basic storybook stories to include a y-offset knob when expanded.

components/src/core/app-bar/app-bar.component.ts
```diff
@@ -0,0 +1,216 @@
...
+                <ng-content select="[pxb-actions]"></ng-content>
+            </mat-toolbar-row>
+        </mat-toolbar>
+    `,
+    host: {
+        class: 'pxb-app-bar',
+    },
+})
+export class AppBarComponent implements OnInit, AfterViewInit, OnChanges, OnDestroy {
+    @ContentChild(ThreeLinerComponent) threeLiner: ThreeLinerComponent;
```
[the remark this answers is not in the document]
emclaug2: You're right, the AppBar doesn't care what the child content is any more, but it doesn't have 3-liner-specific styles to make the transition a built-in feature. 

I've removed the ContentChild reference.

components/src/core/app-bar/app-bar.component.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    @Input() expandedHeight = 200;
```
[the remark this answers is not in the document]
emclaug2: For now I would just stick to pixels.

components/src/core/three-liner/three-liner.component.scss
emclaug2: Good catch; now using rems.

components/src/core/app-bar/app-bar.component.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    private _setNewCollapsedHeight(): void {
+        this.collapsedHeight = this.useDefaultCollapsedHeight
+            ? this._calcDefaultCollapsedHeight()
+            : this.collapsedHeight;
+    }
+
+    private _calcDefaultCollapsedHeight(): number {
+        return document.documentElement.clientWidth <= 600 ? 56 : 64;
```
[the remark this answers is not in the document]
emclaug2: This is a great idea, the defaults have been updated to use rems instead.

emclaug2:
> Did you run through some final checks to make sure everything looks good and works as expected in a showcase-style application?

Yep, hosted here: https://pxb-angular-theme-demo.web.app

emclaug2:
> Just noticing this, should mat-menu-context "text" be right-aligned when in RTL? other then this it all looks good
> 
> image

I would expect this issue to be resolved whenever our RTL storybook plugin handles overlay components' RTL/LTR more gracefully.

docs/ThreeLiner.md
emclaug2: You got it! The link is added now.

docs/ThreeLiner.md
```diff
@@ -1,6 +1,6 @@
 # Three Liner
 
-The `<pxb-three-liner>` can display up to three lines of stylized text or other ng-content. It is most commonly used within the context of a `<pxb-app-bar>` where the text can grow / shrink as the App Bar is expanded and collapsed.
+The `<pxb-three-liner>` can display up to three lines of stylized text or other ng-content. It is most commonly used within the context of a [`<pxb-app-bar>`](./AppBar.md) component where the text can grow / shrink as the App Bar is expanded and collapsed.
```
huayunh, line 3: might break if you are viewing this on storybook

emclaug2: This is working for me, if you click the link here it should redirect to the correct doc: https://pxb-storybook-ng-feature.web.app/?path=/info/components-three-liner--readme

joebochill: That's good, because I think that's how we've been doing cross-linking for a while now :)

components/src/core/app-bar/app-bar.component.ts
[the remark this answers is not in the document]
joebochill: Can we log an enhancement for the future to support other sizes? If somebody wants to override this, they're going to lose the system font size responsiveness.

joebochill (approved):

PR #288, opened 2021-06-10
storybook demo upgrade ng

jeffvg:
- Update storybook demo to use angular 11
- Remove old karma reporter
- Add karma-coverage
- Upgrade pxblue packages
- Add codecov and config
- Update CircleCI to include codecov reporting
- Update angular.json root source to "src"

https://app.codecov.io/gh/pxblue/angular-component-library

image


jeffvg:
storybook demo having issues-

ERROR in ./node_modules/@pxblue/angular-themes/theme.scss
Module build failed (from ./node_modules/postcss-loader/dist/cjs.js):
ValidationError: Invalid options object. PostCSS Loader has been initialized using an options object that does not match the API schema

package.json
```diff
@@ -41,7 +41,11 @@
...
     },
     "devDependencies": {
-        "@pxblue/prettier-config": "^1.0.2",
+        "@pxblue/prettier-config": "^1.0.3",
         "prettier": "^2.2.1"
+    },
+    "dependencies": {
+        "@pxblue/eslint-config": "^2.0.5",
+        "@pxblue/ng-progress-icons": "^1.2.1"
     }
```
joebochill, line 50: ESLint should be in devDependencies. 
Why are we adding progress icons dependency here?

jeffvg (commented):
removed this

package.json
```diff
@@ -6,8 +6,9 @@
     "prettier": "@pxblue/prettier-config",
     "scripts": {
         "initialize": "bash scripts/initializeSubmodule.sh",
-        "test": "cd ./components && yarn test:ci",
+        "test": "cd ./components && yarn test",
```
joebochill, line 9: Is there a reason from switching from `yarn test:ci` (without watcher) to `yarn test` here? I know we originally disabled the watcher because the tests would hang forever in circleCI, but it looks like they passed for this branch, so just wondering.

emclaug2: I'd use `test:ci` in the circleci, and update the command to be `yarn test --no-watch --code-coverage`.  That way the `yarn coverage` command can be removed. 

Other than that, branch looks good to me!

jeffvg: No reason was from testing local and to keep karma window open after tests complete. so its back in

jeffvg: > I'd use `test:ci` in the circleci, and update the command to be `yarn test --no-watch --code-coverage`. That way the `yarn coverage` command can be removed.
> 
> Other than that, branch looks good to me!

not sure what you mean here. Use Test:ci as the call command to execute or do yarn test:ci --code-coverage...
yarn test:ci --code-coverage doesn't work.
We can make it anything but no-watch in CI is the same as test:ci locally which just opens/closes karma

emclaug2: the `yarn test:ci` command should be called only within circle ci, I'm not sure why it's inheriting the `--coverage` and `--no watch` flag instead of just hard-coding those flags directly into the script.  Instead of being able to look at the package.json to see what circle ci is running for the test stage, you have to look at both the .circle config + package.json  to get the full picture of what's happening.

emclaug2: Also for some reason, I thought the coverage script was a duplicated script, but that was my mistake.  The `yarn test:ci` and `yarn coverage` script can coexist.

jeffvg: ok we should be good now. We are calling test:ci for CCI and coverage too

emclaug2 (approved):

joebochill (approved):

PR #290, opened 2021-06-16
Fix info list item wrap prop bug

### LINKED ISSUE #269: InfoListItem wrap prop doesn't break word
#### Describe the bug
If I have a subtitle that is 1 long string and it needs to wrap, it won't.

`wrapTitle` and `wrapSubtitle` should wrap all

1. Go to storybook, `with full config` story
2. Make the subtitle one long word, try to cause an overflow. 

#### Environment

Chrome Windows

doctorbebo:
Fixes #269.

- Fixed info list item wrap prop bug for title and subtitle.

#### To Test:
- Open storybook 
- Turn on wrap title and wrap subtitle props via knobs
- Update title and subtitle to be one long word
- Observe word wrap

### DIFF
--- a/components/src/core/info-list-item/info-list-item.component.scss
+++ b/components/src/core/info-list-item/info-list-item.component.scss
@@ -113,6 +113,7 @@ $denseHeight: 3.25rem;
             white-space: normal;
             overflow: unset;
             text-overflow: unset;
+            overflow-wrap: break-word;
         }
     }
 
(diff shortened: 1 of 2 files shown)


emclaug2 (approved):
Looks good to me! Thanks for fixing this.  

Not sure why the yarn lock changed here, but if it causes any issues I'll just create a new one.

PR #299, opened 2021-06-28
Tweak pixels for app bar story styles

huayunh:
## Changes proposed in this Pull Request:
- Tweaked a few pixels

## Screenshots:

Before:
image
Now: 
image

Before: 
image
Now:
image

### DIFF
--- a/demos/storybook/stories/app-bar/_module.stories.ts
+++ b/demos/storybook/stories/app-bar/_module.stories.ts
@@ -12,7 +12,7 @@ import {
 import { withA11y } from '@storybook/addon-a11y';
 import { withBasicConfig } from './basic-config.stories';
 import * as Colors from '@pxblue/colors';
-import { withModes } from './with-modes.stories';
+import { withVariants } from './with-variants.stories';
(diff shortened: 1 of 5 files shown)


huayunh:
Not sure where this comes from. I checked all the stories and they all have their distinct names.
image

jeffvg:
want to tweak title here....mode went away and is "with variants" ?
image

emclaug2:
> want to tweak title here....mode went away and is "with variants" ?
> image

Yes this makes sense, the `with modes` name was before the API was updated.

huayunh:
Updated. Please verify: https://pxb-storybook-ng-feature.web.app/

emclaug2:
> Not sure where this comes from. I checked all the stories and they all have their distinct names.
> image

In this file, there are 2 DrawerModules defined so that the DrawerLayout stories can have an extra decorator applied.  If you remove the second README story from the 2nd module (line 115) that might fix the issue.

emclaug2 (approved):
Looks good to me, the warning is still there but it's not an error, just a duplicate story warning.

jeffvg (approved):
looks good

PR #311, opened 2021-06-30
Release/v5.0.0

emclaug2:
- Use release/v5.0.0 as the base for all future dev work, upgrade to angular 11


emclaug2:
I need to doublecheck some of these files; this looks wrong.

PR #333, opened 2021-09-09
Fix mobile stepper button uneven width spacing issue

### LINKED ISSUE #332: Stepper dots not centered when buttons are at different size
#### Describe the bug

The dots part of `<pxb-mobile-stepper>` would be off-centered if a button is wider than another.

It should not use flexbox to style this.

image

#### Anything else to add?
This was discovered when I was wondering if we can add a skip button to the right.

emclaug2:
Fixes #332 

- Adds more customization options to the full-config mobile stepper example.
- Ensures that the stepper progress indicator stays centered regardless of  back/next button uneven width.

Check it out here: https://pxb-storybook-ng-feature.web.app/?path=/story/components-mobile-stepper--with-full-config

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,15 @@
 # Change Log
 
 ## v6.0.0 (Unreleased)
+
 ### Added
+
 -   Added new property `unitSpace` to `<pxb-channel-value>` to manage spacing between the value and units.
+
+### Fixed
+
+-   Fixed stepper spacing in `<pxb-mobile-stepper` when Back and Next buttons are uneven width.  
+
 ## v5.0.1 (July 30, 2021)
 
 ### Added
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,15 @@
...
 
 ## v6.0.0 (Unreleased)
+
 ### Added
+
 -   Added new property `unitSpace` to `<pxb-channel-value>` to manage spacing between the value and units.
+
+### Fixed
+
+-   Fixed stepper spacing in `<pxb-mobile-stepper` when Back and Next buttons are uneven width.
```
huayunh, line 11: missing a bracket

huayunh (approved):
Nice!

Can you fix this API table real quick?
image

CHANGELOG.md
emclaug2: You got it!

emclaug2:
Thanks @huayunh once this passes the build I'll merge it in.  Bracket has been added :)

Issue #344, opened 2021-09-20
User Menu `backdropClick` @output not working for bottomsheet


emclaug2:
#### Describe the bug
For bottomsheet overlay, there is no `backdropClick` event emitted.

#### What is the expected behavior?
`backdropClick` should emit an event whenever the backdrop is clicked. 

#### What are the steps to reproduce?
1.  Add a `backdropClick` event onto the `<pxb-user-menu>` and see it doesn't get called for bottomsheet backdrop clicks.

#### Environment

Chrome, Windows
		
#### Anything else to add?
No

emclaug2:
Fixed in #346

PR #346, opened 2021-09-21
Fix #343

### LINKED ISSUE #343: User Menu incorrectly handling bottom sheet state when `open` is manually provided
#### Describe the bug
Bottom sheet does not open whenever the `open` @Input is provided manually.

UserMenu should be smart enough to update state onChanges, instead of only managing state internally. 

1. Checkout this branch: https://github.com/pxblue/angular-design-patterns/tree/feature/account-menu
2.  Open http://localhost:4200/account-menu/in-an-app-bar
3.  Attempt to open the chip example.  This example manages the `open` state outside of the UserMenu component.  

#### Environment

Windows, Chrome		

#### Anything else to add?
Discovered in https://github.com/pxblue/angular-design-patterns/pull/139

emclaug2:
Fixes #343 #344 

- Fix the missing backdrop click event.
- Allow a user to manually control `open` state via @inputs


emclaug2:
An important note about this fix is that the `open` property on the `<pxb-user-menu>` is now required instead of being optional.  I don't think this is a huge issue since v6.0.0 is already a breaking change and all of our documentation & storybook examples already assumed the `open` property was provided.  

To test, I got the mat-chip menu overlay example to work using this version of the library.
https://pxblue-angular-design-patterns.web.app/

emclaug2:
Thinking about this a little more, maybe I can tweak some things to make the `open` prop optional again... Looking into this today.

emclaug2:
I was going to see if there's a way I can refactor this to make `open` not required, but then I realized that a usermenu (most of the time) is going to have clickable items in it that'll close the menu when clicked.

ektasawant:
> An important note about this fix is that the `open` property on the `<pxb-user-menu>` is now required instead of being optional. I don't think this is a huge issue since v6.0.0 is already a breaking change and all of our documentation & storybook examples already assumed the `open` property was provided.
> 
> To test, I got the mat-chip menu overlay example to work using this version of the library.
> https://pxblue-angular-design-patterns.web.app/

Mat-chip inside User Menu is working perfectly fine now.

huayunh:
The functionality looks good. Not an Angular person so I trust you on the code change :P

emclaug2:
I need code owner to check out this branch so we can potentially publish 6.0.0 tomorrow.  @jeffvg @daileytj

daileytj (approved):

PR #349, opened 2021-09-30
Release/6.0.0

emclaug2:
- Merge release branch into dev, getting ready to publish. 

### Added

-   Added new property `unitSpace` to `<pxb-channel-value>` and `<pxb-hero>` to manage spacing between the value and units.

### Changed

-   Changed the `<pxb-user-menu>` component to use a new `UserMenuHeaderComponent` in the menu overlay instead of using a `DrawerHeaderComponent`. 

### Fixed

-   Fixed stepper spacing in `<pxb-mobile-stepper>` when Back and Next buttons are uneven width.  
-   Fixed bug in `<pxb-app-bar>` that prevented class overrides on the root element.
-   Fixed bug in `<pxb-app-bar>` so scroll listeners can no longer attempt to measure `undefined` elements' scroll distance.


emclaug2:
Can be viewed here:  https://pxb-storybook-ng-feature.web.app/

CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
 # Change Log
 
+## v6.0.0 (Unreleased)
```
joebochill, line 3: might as well set the date since this is ready to go

components/src/core/channel-value/channel-value.component.ts
```diff
@@ -37,8 +56,29 @@ export class ChannelValueComponent implements OnChanges {
...
+     * prefixUnitWhitelist: ['$']
+     *
+     * suffixUnitWhitelist: ['%', '℉','°F','℃','°C','°']
+     *
+     * @default auto;
+     * */
+    @Input() unitSpace: UnitSpaceType = 'auto';
 
+    prefixUnitWhitelist = ['$'];
+    suffixUnitWhitelist = ['%', '℉', '°F', '℃', '°C', '°'];
```
joebochill, line 76: did we want to put '/100' in the suffix white list too?

joebochill (approved):
Code looks good to me

components/src/core/channel-value/channel-value.component.ts
joebochill: At this point, I would just log an enhancement for it if we want to

jeffvg (approved):
a lot of good changes here

huayunh:
Plugged the dist into design patterns and everything worked :D

PR #352, opened 2021-10-05
Feature/empty state example

ektasawant:
Fixes PXBLUE-2393 .

- Add new variant `within a card` in storybook.

#### Screenshots:
image

#### To Test:
- http://localhost:6006/?path=/story/components-empty-state--within-a-card


daileytj:
[ ] mat-expansion-panel-header should be 48px when expanded, it's currently 65.
[ ] text within the mat-expansion-panel-header should be subtitle2.
[ ] center align text and keyboard_arrow icon in the mat-expansion-panel-header

daileytj (approved):

PR #358, opened 2021-11-03
Go BLUI

emclaug2:
- This publishes the BLUI component library.

To test out the BLUI components, run `yarn start:storybook`.

I also converted the angular-showcase-demo repo to BLUI.  To test out the BLUI components & themes, checkout the branch below & run `yarn && yarn start`.
https://github.com/pxblue/angular-showcase-demo/tree/feature/blui


jeffvg (approved):
looks good, do we track the  components/README.md refs of pxb later

emclaug2:
> looks good, do we track the components/README.md refs of pxb later

 I'll make a story for this now in JIRA; most pictures in the react & rn component libraries don't need changed but nearly all of angular's will need updated.

andrupu (approved):

daileytj:
Just some things I'm noticing on the showcase that we may want to resolve later or perhaps an audit of this would be nice just to make sure we are using the correct colors as defaults.

These list item tags are hardcoded to have a white background... not sure why but they don't look great on light or dark theme in my humble opinion.
Screen Shot 2021-11-03 at 3 26 45 PM
Screen Shot 2021-11-03 at 3 26 52 PM

Decending character gets cut off of our score card...
Screen Shot 2021-11-03 at 3 27 59 PM

That stop watch icon looks strange to me. Should we be using an outlined variant or something here ¯\_(ツ)_/¯
Screen Shot 2021-11-03 at 3 28 24 PM

None of these things are a direct issue with the component library work as far as I can tell so I'll approve this, but we may want to write a story to double check these colors and other potential issues.

daileytj (approved):

PR #379, opened 2022-01-05
Bug/2723 dynamic appbar el

### LINKED ISSUE #362: Collapsible AppBar Should Handle Scroll Element Changes
#### Describe the bug
This component expects to be fed either a scrollable element id, className or the element itself.  But if you provide an element that is populated `ngAfterViewInit`, the component will not update the scroll element.

Make it so that I can provide a scroll element `ngAfterViewInit`.  

Example:

```
// Parent Component
    scrollEl: Element;

    ngAfterViewInit(): void {
        this.scrollEl = document.getElementsByClassName('mat-sidenav-content')[0];  
        this._ref.detectChanges();
    }

```

In the example above, I want the sidenav content to be the scrollable element, but the app bar component will continue to use the default `undefined. 

...

#### Environment

Windows Chrome
		
#### Anything else to add?
You can work around this by providing the className of the element you want to use instead, it's just confusing that this works and directly passing in the element doesn't.

emclaug2:
Fixes #362 

- Whenever a `scrollContainerClassName`, `scrollContainerId`, or `scrollContainerElement` input changes, stop listening to scroll events and re-initialize the scroll listeners. 

#### To Test
I have a local branch of the showcase that I've used to test and will share.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 
--   Fixed `<blui-app-bar>` not listening for new scroll elements `onChanges`.
+-   Fixed `<blui-app-bar>` not listening for new scroll elements `onChanges` ([#362](https://github.com/brightlayer-ui/angular-component-library/issues/362)).
 
 ## v6.0.1 (December 17, 2021)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,15 @@
 
 -   Added 1rem default padding to `<blui-empty-state>` ([#336](https://github.com/brightlayer-ui/angular-component-library/issues/336)).
 
+### Fixed
+
+-   Fixed `<blui-app-bar>` not listening for new scroll elements `onChanges`.
+
 ## v6.0.1 (December 17, 2021)
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


emclaug2:
I've pushed out a new branch; feature/appbar-dynamic-el in the angular-showcase repo.

Since the branch is difficult to test; here's a gif.  I'll try my best to explain what's happening.

For the first 4 seconds, the "W/ Dynamic Content" container and app bar are paired to each other.   Behavior is normal.
After 4 seconds, I've updated the 2nd Appbar to grow and shrink based on the scroll position of the 1st Appbar container.
After 8 seconds, the 2nd Appbar's scroll element is changed again, back to it's original container.  

app bar gif

daileytj (approved):
Well... that was fun to test! Looks great though... good stuff!

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 -   Added 1rem default padding to `<blui-empty-state>` ([#336](https://github.com/brightlayer-ui/angular-component-library/issues/336)).
 
+### Fixed
+
+-   Fixed `<blui-app-bar>` not listening for new scroll elements `onChanges`.
```
daileytj, line 11: Can you just add a link to the issue, like the entry above, before you merge this?

PR #382, opened 2022-01-07
Force remove the .git folder

emclaug2:
- Force remove the .git folder before attempting to re-init.


jeffvg:
ok this worked for me this time. .git replaced with new .git with current timestamp

jeffvg (approved):

PR #388, opened 2022-01-18
Feature/fix nav item depth v2

### LINKED ISSUE #356: The classe "pxb-drawer-nav-item-depth" is not applied correctly when menu data is not available immediately
#### Describe the bug
In the linked angular project (nav-item-bug.zip), we can see that the classes `pxb-drawer-nav-item-depth-{depth}` are not applied to the nav-items when the data comes from an observable and is not immediately available at the moment a pxb-drawer-nav-group is created. This is explained by the code here: https://github.com/pxblue/angular-component-library/blob/269d8dc892f7ced14806bf3d2f24e45ca1aadc1f/components/src/core/drawer/drawer-body/nav-group/drawer-nav-group.component.ts#L83 

When the data for the menu is updated, the nav-items should be updated correctly

Open the linked project [nav-item-bug.zip](https://github.com/pxblue/angular-component-library/files/7383195/nav-item-bug.zip) and look at the screenshot below

Screenshot 2021-10-20 131844

emclaug2:
Fixes #356 

- Change the way `DrawerNavItemComponents` are initialized so that the `depth` class is applied for async-loaded nav items. 

#### To Test:
- In the submitted issue, there was a demo project provided that had 2 `NavGroups`.  One group had items added synchronously and the other async.  The async-loaded group didn't have a "depth-class" applied to each item.
- https://github.com/emclaug2/async-nav-item-depth has a demo that has been updated to use a local copy of the component library.  Now both `NavGroups` should have the "depth-class" applied.   There are also a few navigation items that are rendered over time; these items should also have the "depth-class" applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 -   Fixed `<blui-app-bar>` not listening for new scroll elements `onChanges` ([#362](https://github.com/brightlayer-ui/angular-component-library/issues/362)).
 -   Fixed `<blui-user-menu>` not rendering bottomsheet content when quickly dismissing and then reopening. ([#345](https://github.com/brightlayer-ui/angular-component-library/issues/345))
 -   Fixed `<blui-drawer-nav-item>` not responding to `expanded` input updates. ([#326](https://github.com/brightlayer-ui/angular-component-library/issues/326))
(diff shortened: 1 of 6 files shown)


jeffvg:
looks good in your repo.


https://user-images.githubusercontent.com/37150565/149989193-1d675ddf-8252-49de-8bbc-518434d0bb66.mov

jeffvg (approved):
looks good, just adjust change log date

PR #390, opened 2022-01-19
Update API docs to include missing components

emclaug2:
- Adding missing docs.


docs/README.md
```diff
@@ -2,12 +2,16 @@
 
 We currently have the following components available for Angular applications:
 
+-   [App Bar](https://github.com/brightlayer-ui/angular-component-library/blob/dev/docs/AppBar.md)
```
huayunh, line 5: might want to point everything to the master branch, or use relative links instead

emclaug2: Just pushed out a fix, it's using 'master' now.  Good catch!

huayunh (approved):

PR #402, opened 2022-02-18
Feature/2919 toolbarmenu

### LINKED ISSUE #374: Add a ToolbarMenu component
This will be a replacement for the existing DropdownToolbar component. We will build a generic component that will handle the dropdown arrow iconography and state and trigger the menu open and closed. The user will supply ng-content for the label (could be text, could be icon + text, etc.) and we will append the dropdown arrow beside their content. Clicking the content will open the menu provided by the user (also supplied via ng content). Most of this functionality can simply be taken out of the existing dropdown toolbar.

Remove the DropdownToolbar component. Make sure that instructions for using the new approach is mentioned in the readme for the Appbar component (e.g., as an example).

Rather than having this functionality built into a Toolbar, we will isolate it into a smaller sub-component that can be used in any toolbar (regular mat toolbar, blui collapsible, etc.)

No

emclaug2:
Fixes #374 

- Add ToolbarMenu

#### To Test:
- Showcase feature/toolbarmenu component


emclaug2:
See it in action here: https://pxb-angular-theme-demo.web.app/blui-components/surface-components

jeffvg (approved):
looks good, dep warning displays in console

daileytj:
Should that menu arrow have a rotate animation or are we cool with the instant flip?

daileytj:
Are there design specs for this that I can use for reference? The body text used in the toolbar looks a little strange to me. Otherwise, looks good.

emclaug2:
> Are there design specs for this that I can use for reference? The body text used in the toolbar looks a little strange to me. Otherwise, looks good.
> 
> And are we planning to update the storybook and showcase? I'm not sure of the scope of this issue so feel free to tell me to kick rocks.

I'll update the showcase as a part of a separate PR, but I won't be able to open it up until we have a beta package submit.  As for the specs, I was looking at the react-component library's implementation as a reference.

And for storybook...
I cannot get the storybook at version 5 to work with angular 13 components, so we have basically kept storybook as-is without any new additions.  We are talking about exploring alternative ways to document our comp libs in the next quarter so temporarily storybook will be unmaintained and eventually replaced.

emclaug2:
> Should that menu arrow have a rotate animation or are we cool with the instant flip?

Ideally it would be a rotation but I just kept all the logic as-is from the dropdown toolbar.  I'll add something in.

daileytj:
@emclaug2 All of that sounds great. I'll approve and if you want to add the animation then that's cool. Long term, it would be nice to have an updated view for this in the component stickersheet.

daileytj (approved):

emclaug2:
I'll make a follow-up PR tomorrow for updating the showcase.

PR #410, opened 2022-03-03
Adjust toolbarmenu line height

emclaug2:
- Fix `toolbarmenu` line height; it should be 1rem in this case.


jeffvg (approved):
aligns better

PR #411, opened 2022-03-03
Bug/2955 empty state border box

### LINKED ISSUE #400: .blui-empty-state-content requires box-sizing: border-box
#### Describe the bug
There's some weird overflow behavior happening within the `EmptyState` component. 

No overflow should be happening. 

Height 100% + 1rem padding = an overflow.

Without Border-Box:
image

With Border-Box:
image

#### Environment

Chrome, Windows
		
#### Anything else to add?
Introduced when we added a default padding of 1rem.

Here's a small example that can be ran in a local .html file. 

```
<div style="background: red; width: 200px; height: 200px; display: flex; justify-content: center; align-items: center">
    <div style="background: blue; height: 100%; width: 50px; padding: 1rem;"></div>
</div>
```

emclaug2:
Fixes #400 

- Add `box-sizing:border-box` to `blui-empty-state-content`. 

This is difficult to test in the showcase app since we don't have an overflow happening there. 
To test, this could be published as a beta package and then integrated my app where the problem was occurring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 -   Updated to Angular 13 for building the library.
 
+### Fixed
+
+-   Fixed container overflow in `<blui-empty-state>` ([#400](https://github.com/brightlayer-ui/angular-component-library/issues/400)).
 ## v6.0.4 (February 14, 2022)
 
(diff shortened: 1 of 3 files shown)


emclaug2:
I've updated the package version to get a beta package published.

emclaug2:
@daileytj @jeffvg The showcase pointer has been updated; give `yarn start:showcase` a try, hopefully it works.  I'm

daileytj (approved):

PR #412, opened 2022-03-03
copy licenses.json to dist

jeffvg:
Fixes # .

- include licenses.json in dist
- 
- 

#### To Test:
-

### DIFF
--- a/scripts/buildComponents.sh
+++ b/scripts/buildComponents.sh
@@ -23,5 +23,6 @@ echo -e "${BLUE}Copying Package Resources${NC}"
 cp -r README.md ./dist/README.md
 cp -r LICENSE ./dist/LICENSE
 cp -r CHANGELOG.md ./dist/CHANGELOG.md
+cp -r components/LICENSES.json ./dist/LICENSES.json
 
 echo -e "${GRAY}Complete${NC}\r\n"


huayunh:
Digging through `node_modules`, I do not see any package with a `LICENSES.json` -alike file as part of their published code. Why do we need this file in our dist? Is it because we are more commercially backed than any other packages we use?

jeffvg:
> Digging through `node_modules`, I do not see any package with a `LICENSES.json` -alike file as part of their published code. Why do we need this file in our dist? Is it because we are more commercially backed than any other packages we use?

@joebochill asked for it to be there so when adopters install our package they will see what we depend on as far as licenses

jeffvg:
@joebochill can you verify this is what's needed

joebochill:
Looks good. @huayunh this is related to requests from the cybersecurity COE to survey our open source license usage. Having this file will make it easy for us to visually identify what our license usage picture looks like. It's not _required_ to be in the dist/npm folder, but having it there will let us have a common location where we know where to find that information.

huayunh (approved):
I see. I think the better approach to that problem is to (also) audit / survey existing packages and ensure that everything we use is "compatible" with what Cyber Security COE calls for — For example, ensure that we do not use packages under GPL licenses

joebochill:
> I see. I think the better approach to that problem is to (also) audit / survey existing packages and ensure that everything we use is "compatible" with what Cyber Security COE calls for — For example, ensure that we do not use packages under GPL licenses

That is exactly what we are doing. The LICENSES file (should) shows all of the licenses for our package dependencies and we'll use that to identify if there are any prohibited packages based on their licenses.

joebochill (approved):

emclaug2 (approved):

PR #438, opened 2022-05-23
Feature/3144 drawer transition

### LINKED ISSUE #82: Fix Drawer Jankiness

emclaug2:
Fixes #82 

- Fixes drawer jankiness when closing drawer.

This bug was happening because the drawer width was being recalculated on `variant` change.

In the `drawer-layout.component.ts` file...

When the variant is updated to `temporary`, the width of the drawer is quickly recalculated to it's full, non-collapsed size.  This is because `isCollapsed()` was returning `false` after the variant was updated to `temporary`.  
I've added in an edge-case that listens for this change & if the drawer is in a collapsed state, avoids the recalculated width until the drawer is fully dismissed from the screen. 

#### Screenshots:
Drawer Dismissable

#### To Test:
- Checkout the showcase project and adjust the size of the screen width until it transitions from a `persistent` to `temporary` variant.  Try this for when the drawer is opened & closed. 

This also should be checked for the `rail` variant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fixed escape key permanently dismissing `<blui-drawer>` ([#426](https://github.com/brightlayer-ui/angular-component-library/issues/426)).
 - Fixed escape key permanently dismissing `<blui-user-menu>`. ([#434](https://github.com/brightlayer-ui/angular-component-library/issues/434)).
+- Fixed `<blui-drawer-layout>` opening side nav when transitioning to a `temporary` state ([#82](https://github.com/brightlayer-ui/angular-component-library/issues/82)).
 
 ## v7.0.1 (April 15, 2022)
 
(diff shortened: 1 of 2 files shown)


emclaug2:
Rail -> Temporary transition can be viewed here: https://github.com/brightlayer-ui/angular-showcase-demo/tree/feature/rail-demo

emclaug2:
I was able to reproduce this & am looking into it now.  Good catch @jeffvg !

jeffvg (approved):
looks good to me, even using rail

PR #452, opened 2022-07-25
Feature/nav item examples

emclaug2:
- Add Nav Item examples page


emclaug2:
This PR is not ready yet, I just want to see the github auto deploy work.

emclaug2:
I'm going to say this PR is ready for review now.  @joebochill @huayunh 

I've added 4 examples for the DrawerNavItem - basic, with icons, with nested items, with selected items.
Can I get some thoughts on:
- How do the descriptions look/read for each example?
- Are the examples doing a good job of showcases the functionalities we're talking about?
- Any examples to add?
- Anything broken?
- Anything else

emclaug2:
Also @daileytj  @jeffvg  I marked as reviewers earlier (before this PR was ready) - re-tagging you since it's ready now.

jeffvg:
Titles & descriptions seem ok. Maybe for example if your doing like a full config, add Nested Drawer Item with divider, nested divider and parent styles...? maybe the hide padding for nested as well...

jeffvg:
also will each example code snippet have a "partial" code view, then show all, hide all, copy all and see on github?

emclaug2:
> also will each example code snippet have a "partial" code view

To make my life easier, there won't be a partial code view right now.  The way these docs are working is that the code snippet & the live-demo are the same thing.  Once we start adding partial code snippets, that'll switch things up & add another level of complexity.

emclaug2:
> also will each example code snippet have a "partial" code view, then show all, hide all, copy all and see on github?

If you meant partial as in it truncates after the first X lines, then yeah we could do something like that.  Ideally the examples here should be small enough that can avoid that though.

jeffvg:
just a general question, looking at the dev doc in figma, we have a mix of examples displaying blui light theme and some no light theme. So should all examples render blui light?

emclaug2:
> Member

I've been skipping the dark theme so 

> just a general question, looking at the dev doc in figma, we have a mix of examples displaying blui light theme and some no light theme. So should all examples render blui light?

I've been skipping the dark theme but so far none of these examples have any colors applied outside of the defaults from our light & dark themes.

jeffvg:
nothing to do with this PR but will be be able to address the drawer jump in/out while on view port size 999 and less while clicking around in examples tab?

emclaug2:
> nothing to do with this PR but will be be able to address the drawer jump in/out while on view port size 999 and less while clicking around in examples tab?

I'll take a look to see if there's an easy way to fix this - this is a side effect of us having a single shared drawer service that manages state in our component library.  Whenever there's multiple drawers active in the same app, they share state & it causes stuff like this to happen.

bkarambe (commented):
I think the tab labels should be fixed, instead of scrolling. Although, I am not sure. @huayunh Can you please confirm that?

https://user-images.githubusercontent.com/25982779/181005536-ca207d22-ca87-4ba8-8cc3-fa2e33212e09.mov

joebochill:
I think the structure of the progressive examples looks good and the level of explanation in each section seems appropriate.

A couple minor items to add:

Be careful with your usage of backtick code blocks...we need to make sure we are setting and sticking to standards here for consistency.

E.g., in the same paragraph you refer to a component using its tag in a code block (I think this is appropriate and aligns with how we've done it in our existing docs), but also in but also by (I think) the name of the component used in a module file
image. Pick one and stick to it, otherwise it's confusing.

In that same vein, when you are using string in code blocks (round, square), you should probably have them in quotes so they aren't confused with variables/Inputs/etc.

joebochill:
> > nothing to do with this PR but will be be able to address the drawer jump in/out while on view port size 999 and less while clicking around in examples tab?
> 
> I'll take a look to see if there's an easy way to fix this - this is a side effect of us having a single shared drawer service that manages state in our component library. Whenever there's multiple drawers active in the same app, they share state & it causes stuff like this to happen.

This will definitely need to be addressed (not necessarily in this PR, but certainly before we launch).

emclaug2:
@bkarambe -  I've pinned the tabs to the top of the page.
@jeffvg & @joebochill -  I've added a patch to subtly patch the shared state.  (This should be fixed in the component library but I don't want to make a patch since it'll take time)
@joebochill  - I've updated the verbiage for the examples

https://blui-angular-docs--pr452-feature-nav-item-exa-8310ts8b.web.app/home

jeffvg:
I didn't notice this before but we can catch this later


https://user-images.githubusercontent.com/37150565/181038356-72ec9708-04fd-41fa-9498-e145d0c4af95.mov

emclaug2:
> I didn't notice this before but we can catch this later
> 
>  NG-dev-doc.mov

This is going to be an issue-worthy observation.  The chevrons in the drawer only show if the drawer is considered open.  Since we are sharing a `DrawerService` between all drawers, the example you see thinks the drawer is closed on small viewports.

jeffvg (approved):
Nav items look good and I did comment earlier that we might want to add something with nested drawer item with divider or nested divider and maybe parent styles.

PR #454, opened 2022-07-26
Add missing drawer routes

emclaug2:
- Add missing drawer routes
- Each page has its correct docs show


jeffvg (approved):
routes look good

PR #455, opened 2022-07-26
Feature/api table styles

emclaug2:
- Make the tables look pretty

### DIFF
--- a/src/themes/doc-light.scss
+++ b/src/themes/doc-light.scss
@@ -39,7 +39,7 @@
                 border-bottom: 1px solid $border;
             }
             thead tr {
-                border-bottom: 1px solid $border!important;
+                border-bottom: 1px solid $border !important;
             }
             thead {
                 background-color: map-get(blui.$blui-white, 100);
--- a/src/themes/doc-light.scss
+++ b/src/themes/doc-light.scss
@@ -9,4 +9,33 @@
     .example-demo-wrapper {
         background-color: map-get(blui.$blui-white, 200);
     }
+    .doc-tab-content {
+        h1,
+        h2,
+        h3 {
+            color: $primary;
+        }
+        h2 {
+            border-bottom: 1px solid rgb(213, 216, 218);
+        }
+        code {
+            background: map-get(blui.$blui-black, 50);
+        }
+        table {
+            * {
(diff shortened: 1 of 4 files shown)


src/themes/doc-light.scss
```diff
@@ -9,4 +9,33 @@
...
+        table {
+            * {
+                box-sizing: border-box;
+            }
+            tr th {
+                text-align: left;
+                margin: 0px;
+                padding: 6px 13px;
+                font-weight: bold;
+                border: 1px solid rgb(213, 216, 218);
```
emclaug2, line 33: Eh, this should probably be a from the colors palette.

huayunh:
There are quite a few places where the markdown did not get interpreted correctly. I don't know if this is within the scope of this PR though

image

emclaug2:
> There are quite a few places where the markdown did not get interpreted correctly. I don't know if this is within the scope of this PR though
> 
> image

I'd say fixing anything table-related can be a part of this story, yes.  But 'primary' 'accent' and 'warn' are strings which is why they appear quoted instead of as a type `boolean` or `undefined`.

I see 'primary' is displayed as a string & a type in 2 separate places, I'll update it to be a string.

emclaug2:
> > There are quite a few places where the markdown did not get interpreted correctly. I don't know if this is within the scope of this PR though
> 
> I'd say fixing anything table-related can be a part of this story, yes. But 'primary' 'accent' and 'warn' are strings which is why they appear quoted instead of as a type `boolean` or `undefined`.
> 
> I see 'primary' is displayed as a string & a type in 2 separate places, I'll update it to be a string.

emclaug2:
@huayunh I just noticed what you pointed out, the table is all wonky - I'll patch that.

huayunh:
If the goal is to make the table look exactly what we specify in our design docs,

image

huayunh:
Other than that, i think your table looks good!

I'd say this thing does not belong in the "API  Docs" tab, as it should be shown under "Example". But that is out of the scope
image

emclaug2:
Thanks @huayunh , PR has been updated.

huayunh (approved):
lgtm 

(I didn't check the code)

PR #458, opened 2022-07-27
Feature/drawer subheader & Footer examples

emclaug2:
- Add DrawerSubheader examples
- Add DrawerFooter examples


huayunh:
I just noticed a thing completely unrelated to the PR: is our drawer header API `title` overriding the native HTML API `title`? Right now if my mouse hovers on a drawer header, it actually shows a tooltip.

image

emclaug2:
> I just noticed a thing completely unrelated to the PR: is our drawer header API `title` overriding the native HTML API `title`? Right now if my mouse hovers on a drawer header, it actually shows a tooltip.

Yes it seems like that's the case.  Alternatively we could do `[title]="'string'"` but it's not as straightforward as `title="string"`.

src/docs/pages/component-docs/components/drawer/drawer-header/examples/with-custom-content.component.ts
```diff
@@ -0,0 +1,20 @@
+import { Component } from '@angular/core';
+
+export const WITH_CUSTOM_CONTENT = `<blui-drawer style="width: 250px">
+    <blui-drawer-header>
+        <div blui-title-content style="display: flex; align-items: start; flex-direction: column">
+               <div class="mat-h4" style="margin-bottom: -8px; margin-top: 8px">Customizable</div>
+               <div class="mat-h2" style="margin: 0">Header Content</div>
+        </div>
```
huayunh, line 8: For the sake of my terminal OCD...

```suggestion
        <div blui-title-content style="display: flex; justify-content: center; flex-direction: column">
               <div class="mat-h4" style="margin-bottom: -8px;">Customizable</div>
               <div class="mat-h2" style="margin: 0">Header Content</div>
        </div>
```

huayunh:
I've also noticed that the line highlighting is not exactly centered — it is shifted upward. Can you bring it down?

image

This margin-top looks better at 1.3rem
image

src/docs/pages/component-docs/components/drawer/drawer-header/examples/with-custom-content.component.ts
emclaug2: All good, it's updated now & will push it out in a bit

emclaug2:
> I've also noticed that the line highlighting is not exactly centered — it is shifted upward. Can you bring it down?

> This margin-top looks better at 1.3rem

This is a little worrysome, I'm seeing it centered on my screen w/ a margin of 1rem.

emclaug2:
@huayunh What browser are you using when you see the code highlighting margin issue?  What about the viewport size when loading?

huayunh:
Google Chrome; any port width (as small as it is allowed to as large as the ultra-wide monitor dare to handle

emclaug2:
> Google Chrome; any port width (as small as it is allowed to as large as the ultra-wide monitor dare to handle

I see... this is going to be a pain to fix.  Google Chrome on Windows look fine, but Google Chrome on Mac has the offset issue.  I'll try to patch this up, but this is the first time I've seen a style difference between the same browser on 2 diff devices. Test both in incognito mode too.

huayunh (commented):
## sub header

Most things look good to me with eyeballing

I will take out the "Sample search functionality" part to keep it compact

I will also leave out the `[hideContentOnCollapse]="false"` in the first example. Our first example should always be a barebone example with minimum properties set up, to avoid confusing our users

## Footer

Do you want to provide this as a more complex example for easy copy-and-paste? This is a variant fairly commonly seen in drawer designs 
image

emclaug2:
> I will also leave out the `[hideContentOnCollapse]="false"` in the first example. Our first example should always be a barebone example with minimum properties set up, to avoid confusing our users

I can remove the `hideContentOnCollapse` after a bug is fixed in the drawer service; this line will allow the content to stay appeared on mobile devices.

> ## Footer
> Do you want to provide this as a more complex example for easy copy-and-paste? This is a variant fairly commonly seen in 

Yep that's fine with me, I'll match the image you've linked.

huayunh:
One other thing I noticed that is outside the PR scope. When the viewport is between 1000px to 1435px, the page body is clipped out, but horizontal scrolls are not allowed. 

image

emclaug2:
@huayunh Removing the `mat-hint` will require us to muddy up the example with more style adjustments, do you still want me to remove it?

Without style adjustments it looks like this:
image

I'm thinking it'll require a negative margin.

emclaug2:
> One other thing I noticed that is outside the PR scope. When the viewport is between 1000px to 1435px, the page body is clipped out, but horizontal scrolls are not allowed.

I won't fix it as part of this issue but I've made an issue for this so we don't forget.
https://github.com/brightlayer-ui/angular-component-library/issues/462

daileytj:
Screen Shot 2022-07-28 at 1 15 15 PM
Screen Shot 2022-07-28 at 1 15 11 PM

Should all of these examples be shown at the 350px width that we use as default for our drawers? Same for the other examples you've already implemented.

emclaug2:
> Should all of these examples be shown at the 350px width that we use as default for our drawer? Same for the other examples you've already implemented.

Great question, if want these examples to be able to render on a mobile device then 350 will be too large; I'm thinking normal iphone size here.

daileytj:
I hear you on the mobile concerns, I was just curious what the most appropriate drop-in code would be for the end-user.

daileytj:
Are we styling the tables at this stage? Looks like the design has vertical lines and more dense rows.

Current Build
Screen Shot 2022-07-28 at 1 35 23 PM

Design
Screen Shot 2022-07-28 at 1 37 58 PM

huayunh:
> Are we styling the tables at this stage? Looks like the design has vertical lines and more dense rows.

I was being lazy in Figma; I didn't want to restyle an entire API table into Material Design's style. What Evan has here for tables are fine

daileytj (approved):
Approving this from the dev side. Anything style was I'll divert to @huayunh

emclaug2:
> Are we styling the tables at this stage? Looks like the design has vertical lines and more dense rows.

 @huayunh approved a separate API table PR yesterday; the designs you're looking at are just copy/pastes from Storybook.  I've updated the tables to look a little more like our design pattern specs.

emclaug2:
@huayunh I've beefed up the drawer subheader demo, it's not quite what you linked to but it's pretty close. 

Beefy Example

huayunh (approved):
Have an organization selected by default, and you will be good to go

PR #467, opened 2022-08-01
Feature/sandbox nav item

emclaug2:
- Add Sandbox for Nav Item

### DIFF
--- a/src/docs/pages/component-docs/shared/scaffold/scaffold.component.scss
+++ b/src/docs/pages/component-docs/shared/scaffold/scaffold.component.scss
@@ -126,7 +126,7 @@
         }
     }
     .props-container {
-        width: 300px;
+        width: 400px;
         box-sizing: border-box;
         min-height: calc(100vh - 64px - 48px - 1px);
         padding: 16px;
--- a/src/docs/pages/component-docs/shared/scaffold/scaffold.component.scss
+++ b/src/docs/pages/component-docs/shared/scaffold/scaffold.component.scss
@@ -118,9 +118,14 @@
     .playground-container {
         display: flex;
         height: 100%;
+        .playground-live-example-wrapper {
+            height: 50%;
+            display: flex;
+            justify-content: center;
+            align-items: center;
+        }
     }
     .props-container {
-        background-color: #e5e5e5;
         width: 300px;
         box-sizing: border-box;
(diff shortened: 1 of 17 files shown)


emclaug2:
@huayunh This PR should be ready for design eyes.  I've made the sandbox full-screen & fixed the overflow issue on the other pages by hiding the drawer at 1280px.

huayunh:
Looks lovely! I like those hint texts.

- I would add a `selected` knob and enable it by default.
- I would also do the hierarchy this way:

image

Other than that, there may be some style updates in the future, but for the first pass this looks good.

emclaug2:
Ok @huayunh I think this is the final version.

- I've added the `selected` knob.
- Adjusted the hint margin margin-left
- Code generation actively ignores default props (if user is using default props, don't show in code snippet)
- Whatever issue I was rambling about in standup this morning (too many props to manage) has been resolved.

huayunh:
I just noticed that the ripple is still on even with the `ripple` knob unchecked

https://user-images.githubusercontent.com/8997218/182656913-896bdc15-7376-4371-82ac-a088f6f9657e.mov

emclaug2:
> I just noticed that the ripple is still on even with the `ripple` knob unchecked

I've made an issue for this one, it's a bug on the DrawerNavItem & not the playground: https://github.com/brightlayer-ui/angular-component-library/issues/468

huayunh (approved):
You have my designer thumbs up! (Didn't check the code)

src/docs/pages/component-docs/shared/scaffold/scaffold.component.scss
```diff
@@ -11,36 +11,44 @@
         width: 100%;
         display: flex;
         justify-content: center;
+        .mat-tab-label {
+            // height: 50px;
+        }
```
daileytj, line 16: What's this doing... or not doing?

daileytj (approved):

src/docs/pages/component-docs/shared/scaffold/scaffold.component.scss
emclaug2: There is still some style cleanup required as part of the overall scaffold - making the bottom Divider of the first NavItem align with the divider under the mat-tab-group was intensional.  I had to tweak the height of the tab group for this stuff to align, but it's still not perfect on mobile devices.

PR #469, opened 2022-08-04
Make it so drawers have their unique instance of state management

### LINKED ISSUE #453: DrawerService needs to differentiate between Drawers
#### Describe the bug
DrawerService is shared between all instances of the Drawer component per app.  This means the `open` or `closed` state is also shared between drawers.  Sometimes 1 drawer can be open and the other can be closed.  

Separate Drawer state between multiple instances of the Drawer component.

Observed here: 
https://github.com/brightlayer-ui/angular-component-library/pull/452#issuecomment-1195585361

#### Environment

Chrome, Google
		
#### Anything else to add?
No

emclaug2:
Fixes #453 

- Makes it so that each time a new `DrawerComponent` is constructed, a new `DrawerState` is made and shared w/ all subsections & `DrawerLayout` parent.
- Removed all `openOnDrawerHover` overrides from doc examples
- Defaults `Drawer` `open` property (required) to true if not required.  This is convenient for our examples, otherwise we'd have to manually mark each drawer as open. 

#### To Test:
- `yarn start` and start hover on Nav Items.  Observe how the main navigation drawer doesn't open/close.


jeffvg (approved):
looks good and issues with drawer are fixed

src/lib/core/drawer/state-listener.component.ts
```diff
@@ -17,11 +55,12 @@ export class StateListener implements OnInit, OnDestroy {
     }
 
     public isOpen(): boolean {
-        return this.drawerService.isDrawerOpen();
+        // console.log(this.drawerService.getNumberOfDrawerInstances());
```
ektasawant, line 58: Please review console.log

ektasawant: Changes looks good only 1 small comment.

ektasawant (approved):

src/lib/core/drawer/state-listener.component.ts
emclaug2: I've removed that extra console log, good catch.

PR #470, opened 2022-08-05
Feature/3386 header playground

emclaug2:
- Add Playground for Drawer Header

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fixed incorrect styling of nested `<mat-drawer-content>` ([#463](https://github.com/brightlayer-ui/angular-component-library/issues/463)).
 - Fixed `<blui-drawer>` sharing a single instance of drawer state ([#453](https://github.com/brightlayer-ui/angular-component-library/issues/453)).
+- Fixed `<blui-drawer-header>` not responding to dynamic icon content. ([#473](https://github.com/brightlayer-ui/angular-component-library/issues/473)).
 
 ## v7.0.2 (June 8, 2022)
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@brightlayer-ui/angular-components",
-    "version": "7.0.3-alpha.1",
+    "version": "7.0.3-alpha.2",
     "description": "Angular components for Brightlayer UI applications",
     "main": "index.js",
     "prettier": "@brightlayer-ui/prettier-config",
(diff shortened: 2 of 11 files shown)


emclaug2:
`color` prop does not work, made an issue for it here: https://github.com/brightlayer-ui/angular-component-library/issues/471

jeffvg (approved):
looks good, covering all options

huayunh:
If an input is optional, please do this for the dropdown, instead of leaving it blank. I have added this to the Figma side.

image

huayunh:
I also have a mixed feeling to the "Show Title Content" option. I am more inclined toward removing it, since we've already shown that on the examples page.

ektasawant:
Changes looks good only agree with https://github.com/brightlayer-ui/angular-component-library/pull/470#issuecomment-1206625297.

emclaug2:
> I also have a mixed feeling to the "Show Title Content" option. I am more inclined toward removing it, since we've already shown that on the examples page.

I'm happy to remove it considering it's on the examples page already.

huayunh:
https://user-images.githubusercontent.com/8997218/183415511-ee6a42ed-f8bf-48a7-bf6a-b720929ea27a.mov

I just noticed this — I am not sure if this is part of the Angular component library or part of this playground.

huayunh:
For the dropdown, I also encourage the use of a divider and have the  "*(Empty)*" option listed on the top.

emclaug2:
> For the dropdown, I also encourage the use of a divider and have the "_(Empty)_" option listed on the top.

I've just pushed out the updated styles for the select menu - it might look a little bit off because the default choice isn't the first option in the list but the divider is there. 👍

huayunh:
that icon issue can still be reproduced this way after a hard reload:
- turn off `Show Icon`
- change to a different color
- turn on `Show Icon`

https://user-images.githubusercontent.com/8997218/183460598-82a38862-452f-4034-b0b9-c73055f5c60b.mov

jeffvg:
I didn't check this since I looked before but is
color prop does not work, made an issue for it here: https://github.com/brightlayer-ui/angular-component-library/issues/471

emclaug2:
> I didn't check this since I looked before but is fixed now? color prop does not work, made an issue for it here: #471

I still see this as broken.

emclaug2:
> that icon issue can still be reproduced this way after a hard reload:
> 
> * turn off `Show Icon`
> * change to a different color
> * turn on `Show Icon`

Thanks @huayunh  - this was an issue with the DrawerHeader component.  I've made an issue for it & patched it since it was a one-liner.

emclaug2:
I'll merge this in now seeing that all the comments have been addressed.  If there's anything else we can address it in the next PR - subheader.

Issue #472, opened 2022-08-08
Standalone Drawer Cannot Open/Close without DrawerLayout wrapper


emclaug2:
#### Describe the bug
Drawer Subheader demo does not transition smoothly between open and closed states. 

#### What is the expected behavior?
Drawer (outside of a DrawerLayout) should be able to open and close correctly. 

#### What are the steps to reproduce?
Drawer Subheader demo does not transition smoothly between open and closed states. 

#### Screenshots or links to minimum reproduction example
```
// drawer-layout.component.scss
    .blui-drawer-content {
        //   transition: none;
        //   width: 100%;
        &.blui-drawer-collapse {
            //      width: 100%;
        }
    }
```

#### Environment

Chrome Windows		

#### Anything else to add?
This may be caused by style bleed from a parent DrawerLayout effecting this standalone example.
Commenting out the lines above seems to solve the issue but I'd like to take a closer look at this during the DrawerLayout & Drawer sandbox stories.

emclaug2:
Fixed in #474

PR #474, opened 2022-08-08
Feature/3387 drawer subheader playground

emclaug2:
- Drawer Subheader Demo & Drawer Footer Demo (they are very similar)
- This demo is not very good right now.  A stand-alone drawer is not able to open & close elegantly.  I've made an issue for this.  I can fix this issue as part of this PR if need be.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fixed incorrect styling of nested `<mat-drawer-content>` ([#463](https://github.com/brightlayer-ui/angular-component-library/issues/463)).
 - Fixed `<blui-drawer>` sharing a single instance of drawer state ([#453](https://github.com/brightlayer-ui/angular-component-library/issues/453)).
 - Fixed `<blui-drawer-header>` not responding to dynamic icon content. ([#473](https://github.com/brightlayer-ui/angular-component-library/issues/473)).
+- Fixed `<blui-drawer-layout>` styles affecting deep nested`<blui-drawer>` children  ([#472](https://github.com/brightlayer-ui/angular-component-library/issues/472)).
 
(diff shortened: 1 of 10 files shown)


huayunh:
Hmm...

1. Can you add a few dummy drawer nav items? It is really hard to see that bottom divider.
2. Is it even possible to make the drawer collapse? We are trying to showcase "hideContentOnCollapse", but the drawer doesn't actually collapse. Although if that makes the drawer overly complicated then we shall scrap that idea.

emclaug2:
> 2. Is it even possible to make the drawer collapse?

This is a little tricky, the answer is yes but it'll involve slightly modifying the component lib.   It would be easier to make this change & test to see if everything is still working if we had the DrawerLayout examples page & sandbox available. 

I can make this change tomorrow morning & get a stand-alone drawer to open/close.

huayunh:
I see. We can also consider not doing the playground at all, given that there are only two inputs and both are fairly straightforward.

emclaug2:
I'd still want to do the playground because it's useful for catching some bugs, like for instance the footer divider goes away whenever the drawer is closed, even with the `hideContentOnCollapse` prop as false.  I'd expect it to be true.  

This sandbox will be a lot more interesting if the drawer can open and close.  I can also add a button for `Open Drawer` in case people don't click the example.

huayunh:
Sounds good. If that becomes too big of a challenge, this is acceptable with me too. The focus here should be about the subheader, not about the drawer state.

image

emclaug2:
I've updated the Footer & Subheader examples to open/close on menu click.  This PR now fixes #472 as well.

huayunh (approved):
Looks good to me. (didn't check the code)

PR #482, opened 2022-08-29
Add drawer playground

emclaug2:
- Add Drawer Playground

### DIFF
--- a/src/docs/pages/component-docs/components/drawer/drawer-footer/drawer-footer-doc.component.ts
+++ b/src/docs/pages/component-docs/components/drawer/drawer-footer/drawer-footer-doc.component.ts
@@ -3,7 +3,7 @@ import { MarkdownService } from 'ngx-markdown';
 import { MarkdownSplitService } from '../../../../../services/markdown-split/markdown-split.service';
 import { BASIC } from './examples/basic.component';
 import { COMPLEX } from './examples/complex.component';
-import { SubheaderPlaygroundKnobs } from '../drawer-subheader/examples/playground.component';
+import { DrawerFooterPlaygroundKnobs } from './examples/playground.component';
 
 @Component({
     selector: 'app-drawer-footer-doc',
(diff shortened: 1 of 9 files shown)


ektasawant:
1. For smaller window size footer is content is breaking
Screenshot 2022-08-30 at 10 33 08 AM
2. `disableActiveItemParentStyles` is not removing styles of parent nav item
Screenshot 2022-08-30 at 10 34 39 AM
3. If I select a value from the slider and after removing the focus slider is still visible. And `openOnHoverDelay` doesn't work until you click on the drawer preview section.
Screenshot 2022-08-30 at 10 35 39 AM
3. In React `openOnHoverDelay` range is from 100 to 1000 and increased by 100ms. Do we need to match?

emclaug2:
@ektasawant  
I've adjusted the slider to match React example's settings.

I've set up the slider so that after sliding it, it doesn't automatically dismiss from the screen; there is still a manual click outside of the slider required to dismiss it.  @huayunh is this behavior acceptable?

`disableActiveItemParentStyles` is working but not immediately - if you change this value it'll require a new item to be selected.

I've fixed the DrawerFooter issue in the Drawer playground.

ektasawant:
> @ektasawant I've adjusted the slider to match React example's settings.
> 
> I've set up the slider so that after sliding it, it doesn't automatically dismiss from the screen; there is still a manual click outside of the slider required to dismiss it. @huayunh is this behavior acceptable?
> 
> `disableActiveItemParentStyles` is working but not immediately - if you change this value it'll require a new item to be selected.
> 
> I've fixed the DrawerFooter issue in the Drawer playground.

working fine 👍

huayunh:
- [x] 1. The current slider works well
- [ ] 2. Can we make drawer nav item dividers to be false by default?
- [ ] 3. Since this playground is meant to show Drawer not Drawer Nav Item... can you make drawer nav item as simple as possible? That is, only icons and titles. No status stripes, no subtitles, (no dividers), less drawer items, only one nav group without title

Question
- Are you planning to make Drawer Layout its own page, or are you going to demonstrate drawer variants in the Drawer playground?

emclaug2:
Thanks for the quick reviews @huayunh ; the Drawer example on the playground has been simplified.

The DrawerLayout is what controls the variants on the angular side so the variants can be found in the DrawerLayout PR.

huayunh (approved):
lgtm (didn't check the code)

One last comment is to organize your properties alphabetically within the same bucket

PR #487, opened 2022-09-06
Feature/3450 app bar examples

emclaug2:
- Add AppBar examples


huayunh:
- [ ] 1. The [design specs](https://brightlayer-ui-dev.github.io/patterns/appbar#design-specifications) for app bars suggest that you would need 32px distance between the hamburger icon and text
- [ ] 2. I encourage you to put the "collapsed" variant on the very top, since that is more commonly seen and is the simplest use case. I am expecting this order: (1) collapsed (2) expanded (3) dynamic/snap (4) custom
- [ ] 3. In "App Bar with Custom Content", consider replacing the three icons on the top right corner with something more meaningful — `search`, `download` and `more_vert` are what I'd suggest

huayunh:
image

huayunh:
Other things lgtm

(Didn't check the code)

src/docs/pages/component-docs/components/app-bar/examples/expanded.component.ts
```diff
@@ -0,0 +1,34 @@
...
+            }
+            .content-body {
+                padding: 16px;
+                background: white;
+            }
+        `,
+    ],
+})
+export class ExpandedComponent {
+    filler = `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas iaculis vulputate arcu, quis vulputate mauris aliquet in. Praesent sollicitudin erat at augue cursus, a scelerisque odio tristique. Aenean quis dui in nisi condimentum viverra eleifend sed odio. Vestibulum a lacinia leo. Proin congue purus vulputate metus vestibulum eleifend. Suspendisse sit amet orci magna. Suspendisse eu odio ut purus ultrices ornare id eu dolor. Cras porttitor semper feugiat. Sed vel ultricies risus, vel dignissim arcu. Vestibulum quis interdum sem. Morbi bibendum tortor ligula, non pellentesque ante ullamcorper sed. Duis ornare felis eget justo faucibus, id tristique felis dignissim. Nullam sapien metus, feugiat egestas convallis eu, vehicula et turpis. Phasellus interdum, justo sit amet maximus pellentesque, eros orci dictum diam, id bibendum risus neque non risus.
```
daileytj, line 31: May want to store this filler text as a constant somewhere. You're using it multiple times already we will likely use this lorem Ipsum text again.

daileytj (approved):
One minor comment, but looks good to me.

emclaug2:
One of the things I'm worried about is duplicating information between the start of the Examples and API docs sections.

Examples Page:
image

API docs Page:
image

huayunh:
Good point... here's what I think. There shall be an introductory paragraph on both the examples page and (very likely) the API docs page:
* Examples page: talks about where it is used; you might be able to extract something from docit. "The app bar is a component typically placed on the top of the page. It shows page title, this and that. It comes with three variants, ..."
* API Docs page: talks about more technical details. "The app bar is inherited from ng mat, but we make it collapsible. It is typically paired with a scroll container and the three-liner component"

How do you feel about this? Or do they still feel redundant to you? The end goal here is so that if a user lands on this page out of nowhere (maybe because a colleague sent them a link), they will still be able to quickly make sense of what they are looking at.

huayunh:
If I have to take something out, I would take out the intro on the API Docs page

emclaug2:
> How do you feel about this? Or do they still feel redundant to you? The end goal here is so that if a user lands on this page out of nowhere (maybe because a colleague sent them a link), they will still be able to quickly make sense of what they are looking at.

Well right now we're in a situation where 1/2 of the components don't have introductory sections so we will have to go back and add them.  The whole reason why I added them to the DrawerLayout page was because all the examples focused on the variants & I thought it would be good to set some context.  

I'm still thinking the intro section is repetitive but if it's a short blurb then it shouldn't be so bad.  If there was a way for both the intro sections to match (examples & api doc), would you consider that?

Either way I'd like to press onward & come back to add / clean up the missing intro sections.  It only takes a couple minutes to add this section per page, but would like to talk about this more maybe in office hours.

huayunh (approved):

PR #489, opened 2022-09-07
Feature/3448 hero examples

emclaug2:
- Hero Examples


huayunh:
- [ ] 1. **Hero with ChannelValue Children.** I would hide the unit space. Or stylize it as `1h 27m` with a space between the hours and the minutes
- [ ] 2. **Hero with Icon Size and Image Types**. Are `bluiWhite` and `bluiBlue` supposed to be some variables imported from somewhere else? 
- [ ] 3. Now that I think about it... I think it might actually be better to take icon colors out, since the goal here is to showcase Heros and HeroBanners, not how to colorize icons.
- [ ] 4. **Hero with Icon Size and Image Types**. There might be a special reason for this, but why is the 3<sup>rd</sup> row of icons referred to as "web icon" instead of "icon font"?
- [ ] 5. **Heroes with a HeroBanner**. Was there a reason to use "divider border" here instead of just turning on a bottom divider using the `divider` property?
- [ ] 6. Doc structure. Similiar to what's commented [here](https://github.com/brightlayer-ui/angular-component-library/pull/487#issuecomment-1239526152).

emclaug2:
> * [ ]  5. **Heroes with a HeroBanner**. Was there a reason to use "divider border" here instead of just turning on a bottom divider using the `divider` property?

I took out the border but the divider prop does not seem to be working.  Funnily enough this is something that we haven't even demoed in our older storybook.

> * [ ]  6. Doc structure. Similiar to what's commented [here](https://github.com/brightlayer-ui/angular-component-library/pull/487#issuecomment-1239526152).

There needs to be some discussion around what the intro section says / looks like.

huayunh (approved):
Can you log an issue for the Hero Banner divider?

emclaug2:
Added a new issue for the divider prop issue: https://github.com/brightlayer-ui/angular-component-library/issues/491

emclaug2:
@huayunh 

Are these the intended styles? 
image

huayunh:
The intended style shall be 
* `1h[space]27m` or just 
* `1h27m`

The 
* `1[space]h27[space]m` 

is off

emclaug2:
> 

Ooof, I was looking at an older version of the deployed site.  Looking good now!

PR #492, opened 2022-09-08
Feature/3456 channel value examples

emclaug2:
- Channel Value Example


huayunh:
# 1

We may want an example to explain the behavior of `unitSpace` — that when it is set to `auto`, certain units are rendered with no space by default.

* `kWh`
  * default
  *  `unitSpace=hide`
* `°C`
  * default
  *  `unitSpace=show`
* `%`
  * default
  *  `unitSpace=show`
* `$` (as a prefix)
  * default
  *  `unitSpace=show`

# 2
Also consider adding a blurb in the API Doc to inform people what these units are.

# 3
Speaking of which, is `unitSpace` a required prop? It is currently listed as required in the API table

emclaug2:
> # 1
> We may want an example to explain the behavior of `unitSpace` — that when it is set to `auto`, certain units are rendered with no space by default.

I'll push out a new example pretty soon that gives a callout to UnitSpace. 

> # 2
> Also consider adding a blurb in the API Doc to inform people what these units are.

I've added some details to the Examples section that talks about the specific units.

> # 3
> Speaking of which, is `unitSpace` a required prop? It is currently listed as required in the API table

Nope, it is not required.  I've updated the API doc.

emclaug2:
> If I navigate to the ChannelValue playground tab... or any playground tab, the tab navigation breaks. 

This is a common theme for all the pages that only that the examples & API section implemented right now.  Next week I can clear up all of these app-breaking empty pages. 

Eh, I've added a `unitSpace` example but I'm not sure if it does exactly as requested.  Take a look!

daileytj:
*Setting unitSpace to 'auto' will show a space for all units except for '%', '℉', '°F', '℃', '°C', and '°'. When prefix is true, the space will be shown for all units except for '$'.

Can you add this blurb below the api table as a note?

daileytj (approved):
I like it! Good stuff. Hopefully @huayunh approves of the updates as well.

emclaug2:
@huayunh I'll merge this in to clear the board and address any follow-up comments as part of the InfoListItem PR that's coming up.

PR #494, opened 2022-09-12
Add InfoListItem examples

emclaug2:
- Add InfoListItem examples & API Page


ektasawant:
Changes look fine 👍 .

ektasawant (approved):

ektasawant:
Not related to this PR but the below links need to update.
Screenshot 2022-09-12 at 6 35 09 PM

https://blui-angular-docs--pr494-feature-3458-infolis-5gju50ok.web.app/Hero.md
Screenshot 2022-09-12 at 6 35 19 PM
https://blui-angular-docs--pr494-feature-3458-infolis-5gju50ok.web.app/ChannelValue.md

emclaug2:
> should the last...more complete example include the full code on github button? also wondering if there are plans to make a component out of the Copy all & Full Code on GitHub. I noticed the order of buttons are different between other examples...

The only piece of the example that isn't available in the code snippet below is the `Colors` variable, but I can refactor this example so that there's no external link needed.

Also yes, the order of these buttons should be consistent across all pages.

emclaug2:
> Not related to this PR but the below links need to update. Screenshot 2022-09-12 at 6 35 09 PM

There's a ChannelValue story for adding a playground - I can updated the doc links as part of that PR.

emclaug2:
@jeffvg I've went through all pages and made the order:

[GITHUB BUTTON] [COPY BUTTON]

We should be good now.

jeffvg (approved):

huayunh:
API page looks good

# 1 Just a thought

the `dense` variant appears a lot in design files. Perhaps we can sneak that in the first example?

image

# 2

This is a default info list item from Figma. It uses `caption` texts for the timestamp. Also note that the time part is bolden instead of `AM`
image

# 3

The third example looks long and complicated. It probably doesn't matter too much since in this case it is on the bottom of the page... but I might take out list item #2 and #5. I would also use a `partial` divider for all list items except for the last one to stay closer to the design convention.

PR #495, opened 2022-09-12
Add InfoListItem playground

emclaug2:
- Add InfoListItem Playground

### DIFF
--- a/src/docs/pages/component-docs/components/info-list-item/examples/playground.component.ts
+++ b/src/docs/pages/component-docs/components/info-list-item/examples/playground.component.ts
@@ -119,10 +119,10 @@ export class PlaygroundComponent implements OnDestroy {
     ${this._playgroundService.addOptionalProp(this.inputs, 'disabled')}
     ${this._playgroundService.addOptionalProp(this.inputs, 'hidePadding')}
     ${this._playgroundService.addOptionalProp(this.inputs, 'iconAlign')}
-    ${this._playgroundService.addOptionalProp(this.inputs, 'statusColor')}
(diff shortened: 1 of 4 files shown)


jeffvg:
looks good, does icon color need to be here?

emclaug2:
> looks good, does icon color need to be here?

Icon color is not a prop on the angular side so I've omitted it - on the React side it'd be present in the playground.

jeffvg:
> 

oh so its removed in angular but adding it? I was just looking at the original
https://brightlayer-ui-components.github.io/angular/?path=/story/components-info-list-item--with-full-config

huayunh:
generated code snippet is missing a bracket `<`

emclaug2:
> > 
> 
> oh so its removed in angular but adding it? I was just looking at the original https://brightlayer-ui-components.github.io/angular/?path=/story/components-info-list-item--with-full-config

On the storybook side we spent some time making all the demos match, even if what we were showcasing wasn't directly related to the component.  This time around we've been sticking to mostly demonstrating the props/inputs.   On the react side it handles colors & bg color as props which is why we've spent the time to add them to the playground.  
We've also not been matching our playgrounds 1-1 between the frameworks for this reason.

jeffvg (approved):
looks good

daileytj (approved):
Looks good to me. 

The `info` input that is defaulted to a string with an ellipse in it caught me up for a second though. It makes me think the text was trimmed at first glance and I expected the `wrapInfo` toggle to show more text until I realized what was going on. I doubt anyone else would stumble there, but wanted to bring it to attention either way.

PR #496, opened 2022-09-12
Add UserMenu playground

emclaug2:
- Add UserMenu Playground

### DIFF
--- a/src/docs/pages/component-docs/components/user-menu/examples/playground.component.ts
+++ b/src/docs/pages/component-docs/components/user-menu/examples/playground.component.ts
@@ -55,7 +55,7 @@ export class PlaygroundComponent implements OnDestroy {
 
     knobListener: Subscription;
     open = true;
-   // positions: ConnectionPositionPair[];
+    // positions: ConnectionPositionPair[];
 
     constructor(private readonly _playgroundService: PlaygroundService) {
         this.knobListener = this._playgroundService.knobChange.subscribe((updatedKnobs: UserMenuPlaygroundKnobs) => {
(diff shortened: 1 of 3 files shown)


jeffvg:
nice!

image

huayunh (approved):
I like how `open` is controlled by both the component and the sidesheet

jeffvg:
any reason to have these included in playground or too much?

image

emclaug2:
> any reason to have these included in playground or too much?

I'll add these in under "Other" props since they're used to create the `positions` prop and not a direct prop.

emclaug2:
Ok so there's a limitation on the angular side; we can add position-knobs so that the demo is updated, but cannot update the generated code.  The `positions` prop requires a new instance of a `ConnectionPositionPair` which can't be done in the template. 

People looking at the codes will have to refer to our code or API docs instead to see _how_ these `originX`, `originY`, etc knobs are used.

huayunh:
hmm... if that's the case, I would leave it out of the playground. I have not run into any situations where someone **has** to change the menu `position`. If you have to showcase how to pass that property / attribute, I recommend doing it on the examples page

emclaug2:
> hmm... if that's the case, I would leave it out of the playground. I have not run into any situations where someone **has** to change the menu `position`. If you have to showcase how to pass that property / attribute, I recommend doing it on the examples page

We've added a Placement example on the Examples page but it's def not interactive.  I'll comment out the placement knobs on the playground page and we can re-add if desired in the future.

jeffvg (approved):
looks good, just remember to comment out the position knobs or remove

PR #497, opened 2022-09-12
Add empty state playground

emclaug2:
-  Add Playground for Empty State


huayunh:
Am I looking at only the playground? It looks like you've got the examples and API docs working as well

huayunh:
# 1 
Examples page, typo 
image

# 2

Is it possible to style that accordion header closer to this in our Angular theme package? Or would we be controlling the style too much?

image

# 3

No complaints to the other two pages

daileytj (approved):
Looks good to me. Just wondering if we should have some options to toggle the icon since it is required and to demonstrate what a different icon may look in the code snippet.

emclaug2:
@huayunh Is the mat-accordion header support to be blue[500] in our theme?  Sounds like a +1 theme issue.

emclaug2:
> @huayunh Is the mat-accordion header support to be blue[500] in our theme? Sounds like a +1 theme issue.

I'll bring this up in the demo.

PR #498, opened 2022-09-13
Feature/3446 mobilestepper playground

emclaug2:
-  Add mobile stepper playground
- The Next & Back Buttons do not do anything; adding this functionality will make the code in the snippet not work anymore.


ektasawant:
Can we update a max range of activeStep prop as per the selection of the step? In the current playground If I select step 3 and activeStep 7 it shows the below output.
Screenshot 2022-09-14 at 11 06 33 AM

huayunh (approved):
Good catch, @ektasawant!

PR #499, opened 2022-09-14
Add channel value playground

emclaug2:
- Add ChannelValue Playground


huayunh:
- [ ] I would not recommend using "12 $" as the default. Maybe `12 kWh` instead?

Everything else looks great!

huayunh:
Ahh I apologize for this. Can you add  a `prefix` knob?

huayunh (approved):

PR #500, opened 2022-09-14
Feature/3440 hero playground

emclaug2:
- Add Hero Playground


huayunh:
There's no `prefix` property for the `Hero` API? 

Everything else looks good

jeffvg (approved):
looks good,see other comment on prefix

emclaug2:
> There's no `prefix` property for the `Hero` API?

Looking at the source it doesn't look there there is.   It's on ChannelValue but not on the Hero.  Sounds like a +1 issue.

emclaug2:
Add a new issue here: https://github.com/brightlayer-ui/angular-component-library/issues/501

PR #535, opened 2022-10-26
Feature/3586 cleanup

### LINKED ISSUE #534: ScoreCard not responding to `badgeOffset` changes correctly
#### Describe the bug
ScoreCard not responding to `badgeOffset` changes correctly; specifically changing the offset value to '0' is causing an issue.

The scorecard component badge offset should respond on all changes.

1. Go to https://brightlayer-ui-components.github.io/angular-dev/components/score-card/playground
2. Set the `badgeOffset` value to 0, not using the slider
  (0 is the default value for badgeOffset; setting the input knob to anything else using keyboard would not reproduce this bug) 

#### Screenshots / Screen recording
badge offset

#### Suggested fix

This is because the logic behind setting the `margin-top` checks if `badgeOffset` is truthy before it is used.  The value 0 is falsey, so it's not used & the default of 'inherit' is used instead.  

#### Anything else to add?
No

emclaug2:
Fixes #534 

- Callout Drawer `width` prop doesn't apply to rail on DrawerLayout playgrond
- Change order of Drawer hierarchy so Group comes before Item
- Add Link on Drawer Nav Group Spacer example
- Adjust Drawer Footer Playground to include Nav Item Icon 
- Reword Hero Icon Formatting description
- Add link to Angular Material List Item on InfoListItem example page
- Reword Mobile Stepper verbiage
- Fix bug in Scorecard (see linked issue)


huayunh:
Just noticed this: what's the reason that Hero Banner is gone from Hero examples?

huayunh:
firebase deploy not triggered this time?

huayunh (approved):
Can confirm everything in the description is fixed

emclaug2:
> firebase deploy not triggered this time?

Looks like I forgot to resolve some merge conflicts; they're fixed now and the auto-deploy is on the way.

emclaug2:
> Just noticed this: what's the reason that Hero Banner is gone from Hero examples?

The HeroBanner doesn't really do much as a component; I opted to remove it entirely but it's still used in the first example on Hero examples page.

emclaug2:
Also one thing I forgot the mention in the description - 

Code highlighting randomly disappears from the app, but I found the cause and this won't happen anymore. 
To recreate the issue from the dev branch, navigate to a README tab & refresh the page.  Then switch tabs back to Examples, observe no code-highlighting.

huayunh:
> The HeroBanner doesn't really do much as a component; I opted to remove it entirely but it's still used in the first example on the Hero examples page.

hmm... feels weird to use a single hero in a single hero banner. Consider removing it. I saw that the hero banner is also used in "Icon Formatting" — that should be enough

jeffvg (approved):
looks good

PR #592, opened 2023-03-02
556 Channel value issue with zero value

### LINKED ISSUE #556: Channel value issue with zero value
#### Describe the bug

If you pass 0 as a value for the channel value component, it won't display the value.

The 0 value is visible

Pass 0 has value to the component (see code snippet below).

#### Screenshots / Screen recording

channel-value

#### Code snippet / Link to minimum reproduction example
```
<blui-channel-value
  [value]="0"
  [units]="'kWh'"
>
</blui-channel-value>

```

#### Your environment information
```
    "@brightlayer-ui/angular-components": "7.0.3",
    "@brightlayer-ui/angular-themes": "7.0.1",
    "@brightlayer-ui/chartjs": "^1.0.0",
    "@brightlayer-ui/colors": "3.1.1",
    "@brightlayer-ui/icons": "1.11.0",
    "@brightlayer-ui/ng-progress-icons": "1.2.1",
```

#### Suggested fix

My assumption is that there is a if condition that check the value but 0 will return false therefore hiding the value.

raghavenc5:
Fixes #556 .

- Check the input value type as string or number then convert into string if number as 0
- 
- 

-
image

#### To Test:
- yarn start

-

### DIFF
--- a/src/lib/core/channel-value/channel-value.component.ts
+++ b/src/lib/core/channel-value/channel-value.component.ts
@@ -2,7 +2,6 @@ import { ChangeDetectionStrategy, Component, Input, OnChanges, ViewEncapsulation
 import { requireInput } from '../../utils/utils';
 
 export type UnitSpaceType = 'show' | 'hide' | 'auto';
-//console.log(value);
 /** [ChannelValue Component](https://brightlayer-ui-components.github.io/angular/?path=/info/components-channel-value--readme)
  *
  * Displays a value/unit combination.
@@ -29,7 +28,7 @@ export type UnitSpaceType = 'show' | 'hide' | 'auto';
                 class="blui-channel-value-value"
                 [class.blui-channel-value-remove-space]="unitSpace === 'hide'"
             >
-                {{ value }}
+                {{ getValueAsString() }}
             </div>
             <div
                 *ngIf="units && !prefix"
@@ -54,14 +53,7 @@ export class ChannelValueComponent implements OnChanges {
     /** The text to display for the units (light text) */
     @Input() units: string;
     /** The text to display for the value (bold text)*/
-    @Input('value')
-    get value(): string | number {
-        return this._value;
-    }
-    set value(newValue: string | number) {
-        this._updateRenderedValue(newValue);
-    }
-    private _value: string | number;
+    @Input() value: string | number;
     /** Text color */
     @Input() color: string;
     /** Show a space between the value and units. Can be `show` | `hide` | `auto`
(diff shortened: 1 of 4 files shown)


src/lib/core/channel-value/channel-value.component.ts
```diff
@@ -2,6 +2,7 @@ import { ChangeDetectionStrategy, Component, Input, OnChanges, ViewEncapsulation
 import { requireInput } from '../../utils/utils';
 
 export type UnitSpaceType = 'show' | 'hide' | 'auto';
+//console.log(value);
```
daileytj, line 5: Let's make sure we clean up our console logs

src/lib/core/channel-value/channel-value.component.ts
```diff
@@ -81,4 +89,12 @@ export class ChannelValueComponent implements OnChanges {
...
             : this.suffixUnitWhitelist.includes(this.units);
     }
+
+    /** Update the text of the Channel value element in the DOM, creating the element if necessary. */
+    private _updateRenderedValue(newValue: string | number): void {
+        const newValueNormalized: any = `${newValue ?? ''}`.trim();
+
+        // Check the value as interger and convert into the string value
+        this._value = !isNaN(Number(newValueNormalized)) ? String(newValueNormalized) : newValueNormalized;
+    }
```
daileytj, line 99: This feels like overkill. I would have gone with a simpler approach where you render `{{ getValueAsString() }}` instead of `{{ value }}` in the component template and then use a simple function like this to return the value as a string: 

`getValueAsString(): string {
        return String(this.value);
    }`

If we wanted to make it generic we could take in a value for the getValueAsString function and convert that instead. Feel free to state your case on why your approach is preferred.

src/lib/core/channel-value/channel-value.component.ts
```diff
@@ -53,7 +54,14 @@ export class ChannelValueComponent implements OnChanges {
...
     /** The text to display for the value (bold text)*/
-    @Input() value: string | number;
+    @Input('value')
+    get value(): string | number {
+        return this._value;
+    }
+    set value(newValue: string | number) {
+        this._updateRenderedValue(newValue);
+    }
+    private _value: string | number;
```
daileytj, line 64: We should be able to achieve this fix without the need for a getter and setter on that input.

daileytj (commented):
As Suraj mentioned, you may want to check your prettier config or run `yarn prettier` prior to making commits so they are in line with our code-standards. Great job adding a test case for this!

daileytj (approved):

PR #594, opened 2023-03-06
Bug/515 toolbar menu icon centered

### LINKED ISSUE #515: Toolbar Menu, left icon should be vertically centered
ToolbarMenu icon should be center-aligned.

It's not centered. 

#### How to recreate?
1. Navigate to the ToolbarMenu example that showcases an icon, [here](https://brightlayer-ui-components.github.io/angular/components/toolbar-menu/examples). 
2. Observe that the icon is not vertically center-aligned. 

No

#### Anything else to add? 
Center this icon
image

raghavenc5:
Fixes #515  .

-  Updated the line-height in the .blui-toolbar-icon-wrapper .mat-icon CSS class
- 
- 

-
image

#### To Test:
-  yarn test

-

### DIFF
--- a/src/lib/core/toolbar-menu/toolbar-menu.component.scss
+++ b/src/lib/core/toolbar-menu/toolbar-menu.component.scss
@@ -17,13 +17,16 @@
         align-items: center;
         cursor: pointer;
     }
+    .blui-toolbar-icon-wrapper {
+        display: flex;
+        align-items: center;
+    }
     .blui-toolbar-menu-label {
-        line-height: 1rem;
+        line-height: inherit;
     }
     .blui-toolbar-icon-wrapper .mat-icon {
         font-size: 1rem;
         width: 1rem;
-        line-height: 1.3rem;
         height: 1rem;
         margin-right: 8px;
     }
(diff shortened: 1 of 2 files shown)


src/lib/core/toolbar-menu/toolbar-menu.component.scss
```diff
@@ -23,7 +23,7 @@
     .blui-toolbar-icon-wrapper .mat-icon {
         font-size: 1rem;
         width: 1rem;
-        line-height: 1rem;
+        line-height: 1.3rem;
```
daileytj, line 26: How did you calculate this value?

It seems really close to being centered, but I'm curious about the exact math here to ensure it's centered.

Here are some screenshots with center lines added. It's difficult to tell exactly... could you take a look @huayunh and let me know how to test something like this in the future?
image

huayunh:
I have not spun the project on my end yet, but here's what I expect

### 1
- Both the leading icon and the trailing icon shall be 24px
  - Ideally, these two icons shall be sized the same size as the `line-height` (different from `font-size`), but I do not know if this is technically feasible. Might want to test it out. Like try setting `line-height` to be `inherit` might help

### 2
image

src/lib/core/toolbar-menu/toolbar-menu.component.scss
huayunh: you are doing what I would do. Center lining... inspector... My eyes can detect as low as a 1px shift, and when I eyeballed the screenshot in the description it looked good to me.

But I would try to avoid magic numbers like these unless absolutely needed

daileytj (approved):
Great work @raghavenc5!

PR #611, opened 2023-05-01
blui 3952 GA4 trigger deploy with updated id

JeffGreiner-eaton:
Fixes BLUI-3952.

- Test if existing GA works with updated ID
- 
- 

-

#### To Test:
- 

-

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Brightlayer UI Angular Components
+trigger deploy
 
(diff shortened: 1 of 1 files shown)


JeffGreiner-eaton:
> Looks good. Will the react updates be as simple as swapping out the "ga" code? If not, that might be worth exploring as part of this spike if you have the bandwidth.

so far yes and react dev docs was done last month.

